PR #2, opened 2017-04-22
Windows virtualenv paths.

markrwilliams:


mahmoud:
Hmmm, isn't the point of this PR to increas coverage? As in, the codecov report ought to say that a change would occur?

markrwilliams:
Indeed.  Don't merge it until it does :)

mahmoud:
I'm seeing a 2% improvement, looks like it's good to go. Thanks!

PR #12, opened 2017-06-20
Integrating the scheme registration subsystem

mahmoud:
This PR covers hyperlink's method for handling of schemes without default support. 

As you may know, the URL scheme (e.g., http/ssh/mailto/etc.) controls some pretty important semantics for later parts of the URL. I've tried to make the list of built-in schemes as complete as possible, with ~60 schemes supported by default. 

That said, the list is always growing, and I didn't want to make the same mistake as the standard library's urlparse, by hardcoding several lists without some developer recourse.

Even though this PR only adds tests, I'd be very happy to get any feedback on the system as a whole.

### DIFF
--- a/hyperlink/test/common.py
+++ b/hyperlink/test/common.py
@@ -1,0 +1,70 @@
+
+
+from unittest import TestCase
+
+
+class HyperlinkTestCase(TestCase):
+    """This type mostly exists to provide a backwards-compatible
+    assertRaises method for Python 2.6 testing.
+    """
+    def assertRaises(self, excClass, callableObj=None, *args, **kwargs):
+        """Fail unless an exception of class excClass is raised
+           by callableObj when invoked with arguments args and keyword
+           arguments kwargs. If a different type of exception is
+           raised, it will not be caught, and the test case will be
+           deemed to have suffered an error, exactly as for an
+           unexpected exception.
+
+           If called with callableObj omitted or None, will return a
+           context object used like this::
+
+                with self.assertRaises(SomeException):
+                    do_something()
+
+           The context manager keeps a reference to the exception as
+           the 'exception' attribute. This allows you to inspect the
+           exception after the assertion::
+
+               with self.assertRaises(SomeException) as cm:
+                   do_something()
+               the_exception = cm.exception
+               self.assertEqual(the_exception.error_code, 3)
+        """
+        context = _AssertRaisesContext(excClass, self)
+        if callableObj is None:
+            return context
+        with context:
+            callableObj(*args, **kwargs)
+
+
+class _AssertRaisesContext(object):
+    "A context manager used to implement HyperlinkTestCase.assertRaises."
+
+    def __init__(self, expected, test_case, expected_regexp=None):
+        self.expected = expected
+        self.failureException = test_case.failureException
+        self.expected_regexp = expected_regexp
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if exc_type is None:
+            try:
+                exc_name = self.expected.__name__
+            except AttributeError:
+                exc_name = str(self.expected)
+            raise self.failureException("%s not raised" % exc_name)
+        if not issubclass(exc_type, self.expected):
+            # let unexpected exceptions pass through
+            return False
+        self.exception = exc_value  # store for later retrieval
+        if self.expected_regexp is None:
+            return True
+
+        expected_regexp = self.expected_regexp
+        exc_str = str(exc_value)
+        if not expected_regexp.search(exc_str):
+            raise self.failureException('%r does not match %r' %
+                                        (expected_regexp.pattern, exc_str))
+        return True
--- a/hyperlink/test/common.py
+++ b/hyperlink/test/common.py
@@ -40,31 +40,19 @@ class HyperlinkTestCase(TestCase):
 class _AssertRaisesContext(object):
     "A context manager used to implement HyperlinkTestCase.assertRaises."
 
-    def __init__(self, expected, test_case, expected_regexp=None):
+    def __init__(self, expected, test_case):
         self.expected = expected
         self.failureException = test_case.failureException
-        self.expected_regexp = expected_regexp
 
     def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, tb):
         if exc_type is None:
-            try:
-                exc_name = self.expected.__name__
-            except AttributeError:
-                exc_name = str(self.expected)
-            raise self.failureException("%s not raised" % exc_name)
+            exc_name = self.expected.__name__
+            raise self.failureException("%s not raised" % (exc_name,))
         if not issubclass(exc_type, self.expected):
             # let unexpected exceptions pass through
             return False
         self.exception = exc_value  # store for later retrieval
-        if self.expected_regexp is None:
-            return True
-
-        expected_regexp = self.expected_regexp
-        exc_str = str(exc_value)
-        if not expected_regexp.search(exc_str):
-            raise self.failureException('%r does not match %r' %
-                                        (expected_regexp.pattern, exc_str))
         return True
--- a/hyperlink/test/test_common.py
+++ b/hyperlink/test/test_common.py
@@ -1,0 +1,106 @@
+"""
+Tests for hyperlink.test.common
+"""
+from unittest import TestCase
+from .common import HyperlinkTestCase
+
+
+class _ExpectedException(Exception):
+    """An exception used to test HyperlinkTestCase.assertRaises.
+
+    """
+
+
+class _UnexpectedException(Exception):
+    """An exception used to test HyperlinkTestCase.assertRaises.
+
+    """
+
+
+class TestHyperlink(TestCase):
+    """Tests for HyperlinkTestCase"""
+
+    def setUp(self):
+        self.hyperlink_test = HyperlinkTestCase("run")
+
+    def test_assertRaisesWithCallable(self):
+        """HyperlinkTestCase.assertRaises does not raise an AssertionError
+        when given a callable that, when called with the provided
+        arguments, raises the expected exception.
+
+        """
+        called_with = []
+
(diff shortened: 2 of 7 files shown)


hyperlink/test/common.py
```diff
@@ -0,0 +1,70 @@
+
+
+from unittest import TestCase
+
+
+class HyperlinkTestCase(TestCase):
```
Lukasa: Intended only as a comment: would it be better to use py.test or add a conditional dependency on unittest2 than directly backport this code?

hyperlink/test/test_scheme_registration.py
```diff
@@ -0,0 +1,43 @@
...
+        # test netloc works even when the original gives no indication
+        u2 = URL.from_text('deltron:')
+        u2 = u2.replace(host='example.com')
+        assert u2.to_text() == 'deltron://example.com'
+
+        # test default port means no emission
+        u3 = URL.from_text('deltron://example.com:3030')
+        assert u3.to_text() == 'deltron://example.com'
+
+        register_scheme('nonetron', default_port=3031)
```
Lukasa: I assume this shouldn't be here?

hyperlink/test/common.py
Lukasa: Actually, this is really only a Python 2.6 compatibility concern. Do we need to maintain Python 2.6 compat?

mahmoud: I'd love to use py.test directly. This is more an artifact of Twisted's approach/Moshe's recommendation that I write in unittest and run with a more modern test runner like trial/py.test.

I figured I'd handle that as a separate refactor, separate the concerns a bit.

As far as 2.6 compat, well, it's easy. The code itself needed no changes, only the tests. And as someone who's working days porting a codebase off 2.6, I certainly appreciate it. :)

hyperlink/test/test_scheme_registration.py
mahmoud: Ah, that last line may need an assert of some form. The lines above test the invariant that default ports aren't emitted.

hyperlink/test/common.py
Lukasa: I mean, I guess the question is only: do we believe 2.6 support is necessary? I suspect the answer to that question boils down to "do we support a version of Twisted that supports 2.6", but if you think 2.6 support is minimal effort then I'm not going to push you to drop it. :wink:

hyperlink/test/test_scheme_registration.py
Lukasa: Ah, ok. Either a comment or an assert would be good. =)

mahmoud: Added a couple extra asserts.

mahmoud:
Note that @markrwilliams said he'd take a look at the codecov. Not sure what the recommended practice is for coverage on code in the test package itself, but he'll at least figure out why the proper XML isn't being sent to codecov.

glyph:
> Not sure what the recommended practice is for coverage on code in the test package itself

My general policy is that if the code isn't covered, you can probably delete it ;-).

This itself is a backport, which already has tests, which just aren't in this package; in this very specific case, I'd put the portions of the backport which are a literal copy/paste out of another codebase into a module which then gets explicitly excluded from coverage.

hyperlink/test/common.py
markrwilliams: I'd prefer switching to pytest, since this backport runs on _all_ versions of Python.  `_AssertRaisesContext` is not exactly the same as [what's in the standard library](https://github.com/python/cpython/blob/2.7/Lib/unittest/case.py#L109-L128).  I think the differences are irrelevant but it's not a great situation to be in.

But that's for a future PR of course.

hyperlink/test/test_common.py
```diff
@@ -0,0 +1,106 @@
...
+
+        try:
+            self.hyperlink_test.assertRaises(_ExpectedException,
+                                             doesNotRaiseExpected)
+        except _UnexpectedException:
+            pass
+
+    def test_assertRaisesWithCallableDoesNotRaise(self):
+        """HyperlinkTestCase.assertRaises raises an AssertionError when given
+        a callable that, when called, does not raise any exception.x
```
markrwilliams: Trailing x

hyperlink/test/test_common.py
```diff
@@ -0,0 +1,106 @@
...
+        """HyperlinkTestCase.assertRaises does not raise an AssertionError
+        when used as a context manager with a suite that raises the
+        expected exception.  The context manager stores the exception
+        instance under its `exception` instance variable.
+
+        """
+        with self.hyperlink_test.assertRaises(_ExpectedException) as cm:
+            raise _ExpectedException
+
+        self.assertTrue(isinstance(cm.exception, _ExpectedException))
```
markrwilliams: Python 2.6's `unittest.TestCase` doesn't have `assertIsInstance`.  Amazing.

hyperlink/test/test_scheme_registration.py
```diff
@@ -0,0 +1,51 @@
...
+from .._url import register_scheme, URL
+
+
+class TestSchemeRegistration(HyperlinkTestCase):
+
+    def test_register_scheme_basic(self):
+        register_scheme('deltron', uses_netloc=True, default_port=3030)
+
+        u1 = URL.from_text('deltron://example.com')
+        assert u1.scheme == 'deltron'
```
markrwilliams: If this is a `unittest.TestCase`, these should be `self.assertEquals`

hyperlink/test/test_scheme_registration.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+from .common import HyperlinkTestCase
+from .._url import register_scheme, URL
+
+
+class TestSchemeRegistration(HyperlinkTestCase):
+
+    def test_register_scheme_basic(self):
+        register_scheme('deltron', uses_netloc=True, default_port=3030)
```
markrwilliams: This modifies [global state inside `hyperlink`](https://github.com/python-hyper/hyperlink/blob/master/hyperlink/_url.py#L328-L334), but it doesn't look like there's anything that will unwind this change after this test's completion.  That's a recipe for trouble down the way.

It looks like a `setUp` method could copy `hyperlink._url.SCHEME_PORT_MAP` and `hyperlink._url.NO_NETLOC_SCHEMES`, which a `tearDown` method could restore.  Please add something like this to minimize the chances of unexpected dependencies between tests.

markrwilliams: Also, what happens if I do:

```python
register_scheme("something")
```

And `uses_netloc` remains `None`?  [It looks like nothing!](https://github.com/python-hyper/hyperlink/blob/master/hyperlink/_url.py#L335-L336)  If that's the case, `uses_netloc` should be a required argument.

hyperlink/test/test_scheme_registration.py
```diff
@@ -0,0 +1,51 @@
...
+        u4 = u4.replace(host='example.com')
+        assert u4.to_text() == 'nonetron://example.com'
+
+        register_scheme('noloctron', uses_netloc=False)
+        u4 = URL(scheme='noloctron')
+        u4 = u4.replace(path=("example", "path"))
+        assert u4.to_text() == 'noloctron:example/path'
+
+    def test_register_no_netloc_scheme(self):
+        register_scheme('netlocless', uses_netloc=False)
```
markrwilliams: Please move the assertions and code in `test_register_scheme_basic` that has `uses_netloc=False` (lines 33-36) here.

tox.ini
```diff
@@ -13,3 +13,4 @@ changedir = .tox
 deps = coverage
 commands = coverage combine --rcfile {toxinidir}/.tox-coveragerc
            coverage report --rcfile {toxinidir}/.tox-coveragerc
+           coverage html --rcfile {toxinidir}/.tox-coveragerc -d {toxinidir}/htmlcov
```
markrwilliams: This is a little expensive to generate so it may be worth putting this in a separate tox env.  On the other hand, it's really nice that I can now rely on `htmlcov` hanging around.  It's totally OK if this stays but something to think about?

hyperlink/test/test_common.py
mahmoud: emacs'd (fixed)

hyperlink/test/test_common.py
mahmoud: amazing what the stdlib can get away with

hyperlink/test/test_scheme_registration.py
mahmoud: I think what should be happening is that it should default to the truthiness of default_port. Can't have a default port if you don't use network locations.

hyperlink/test/test_scheme_registration.py
mahmoud: Per an earlier discussion, the TestCase element is an artifact of leaving the old twisted.python.url test cases as stable as can be. I'm planning a pytest migration very soon, once other aspects settle down. These tests will port right over.

hyperlink/test/test_scheme_registration.py
markrwilliams: So it should be `use_netloc=False`?

mahmoud: Looking at the sheer prevalence of use, I think use_netloc=True is a more sensible default and would meet more expectations. I'll remove the option of having `None`, it's not saving anyone a boatload of typing or anything. Sound good?

markrwilliams: Defaulting to `uses_netloc=True` puts you in the uncomfortable position of having a default signature whose invocation raises an exception, because `default_port=None`.  If that seems OK, I'm OK with this.

But why not just make `uses_netloc` a required argument?  Then `default_port` can remain `None`, because if the user chooses to do `register_scheme('deltron', uses_netloc=False)`, they won't encounter an exception.

mahmoud:
Alright, all changes for this PR address, merging per @markrwilliams' instruction. Thanks again all!

PR #13, opened 2017-06-24
Remove a duplicate word

mineo:


mahmoud:
The echo was for effect! (Joking, thank you for the good catch!) :)

Issue #21, opened 2017-07-02
to_unicode() is harmful


wsanchez:
I've had some version of this bug more than once:

```
>>> from hyperlink import URL
>>> URL.from_text(b"/foo")
URL.from_text("b'/foo'")
```

Note the `bytes` were interpreted as `"b'/foo'"` instead of the (almost certainly) more-expected `"/foo"`.

This is because `URL.from_text(s)` calls `to_unicode(s)` which calls `unicode(s)`.

This "works" in Python 2 and does the above in Python 3.

It seems to me that raising an exception insofar as `from_text` should be given text.

wsanchez:
This is especially problematic in that it may be common (`request` is a Twisted `IRequest`, so `request.uri` is a `bytes`):

```
requestURL = URL.fromText(request.uri)
```

mahmoud:
Yeah, I noticed this on the last issue and agree. Should be a quick change I can do right now and I'll assign you as reviewer.

mahmoud:
I'm gonna close this one in favor of #20, since upon further review that basically describes the exact same problem (just slightly more bytes-specific). Thanks for the report!

Issue #20, opened 2017-06-27
URL takes any type as input


wsanchez:
URL takes any type as input.   This often is wrong, most notably with `bytes` in Python 3:

```
>>> URL.fromText(b"/foo").asText()
"b'/foo'"
>>> URL.fromText(object()).asText()
'%3Cobject%20object%20at%200x100538160%3E'
```

Raising `TypeError` seems more useful in these cases.

mahmoud:
Fixed in 17.3.0, just released, thanks again!

PR #22, opened 2017-07-02
enforce decoding prior to calling from_text

mahmoud:
Pretty straightforward changes to require that `from_text` is only called with decoded strings (`unicode` objects on py2, etc.), per the discussion in #21 and #20.


mahmoud:
@wsanchez feel free to take a look and give me an LGTM if you've got a moment. (wrt the codecov check, this PR is only removing code, codecov is just being a nuisance atm. I'll add a codecov config separately to tweak its behavior.)

wsanchez:
Looks great, thanks.

Issue #16, opened 2017-06-25
reserved characters are treated inconsistently and not sensibly preserved


glyph:
This has been a design flaw since the inception of the library, so, mea culpa on that.

Fundamentally, preserving, escaping, and encoding "reserved" characters is entirely the URL object's job, and it's failing at that.  Possibly the most succinct demonstration of the problem is this:

```python
>>> u = URL()
>>> u = u.child(u'/')
>>> u = u.asIRI()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    u = u.asIRI()
  File
"/Users/glyph/Library/Python/2.7/lib/python/site-packages/hyperlink/_url.py",
line 1116, in to_iri
    fragment=_percent_decode(self.fragment))
  File
"/Users/glyph/Library/Python/2.7/lib/python/site-packages/hyperlink/_url.py",
line 861, in replace
    userinfo=_optional(userinfo, self.userinfo),
  File
"/Users/glyph/Library/Python/2.7/lib/python/site-packages/hyperlink/_url.py",
line 606, in __init__
    for segment in path))
  File
"/Users/glyph/Library/Python/2.7/lib/python/site-packages/hyperlink/_url.py",
line 606, in <genexpr>
    for segment in path))
  File
"/Users/glyph/Library/Python/2.7/lib/python/site-packages/hyperlink/_url.py",
line 410, in _textcheck
    % (''.join(delims), name, value))
ValueError: one or more reserved delimiters /?# present in path segment: u'/'
>>>
```

This is - obviously I hope - the wrong place to be failing with an error like this.

There was previously *some* attempt to preserve these characters in the data model and escape them only upon stringification, but d26814c074c6f9787e62af907df17fbd68fde615 wrecked these semantics.  (In fairness: the attempt to do this was broken, and there are some places, like the scheme, where certain characters indeed *cannot* be represented, so this direction isn't entirely wrong.)

Fundamentally if a user wants to encode slashes, question marks, hash signs or whatever else that a human might, for example, type into a text field, then it should be possible to do that.

We could fix this obvious manifestation of the problem by just putting back the escape-only-on-`asText` logic, but that still leaves an even more pernicious problem:

```python
>>> u = URL(path=tuple([u'%2525']))
>>> u.asText()
u'%2525'
>>> u.asIRI().asText()
u'%25'
>>> u.asIRI().asIRI().asText()
u'%'
>>> 
```

[...]

PR #23, opened 2017-07-03
Fix percent decoding and normalization

mahmoud:
Per #16, hyperlink's inherited behavior was applying percent-decoding too broadly. Each percent-encodable field in the URL (userinfo, path, query string, and fragment), has certain special characters that should *never* be decoded [1].

To this end, this PR adds and utilizes a `_decode_XXX_part()` function for each of the fields, symmetrical to the `_encode_XXX_part()` that was already there.

Several test cases were added, as well as a general test for the invariant that multiple calls of `to_iri()` will not generate different URL objects.

@glyph has shown the most interest in this feature, so I've added him as reviewer.

[1]: To take the path as an example, the special characters include delimiters, such as the slash (`/`), question mark (`?`), and hash mark (`#`), as well as the percent sign (`%`) itself.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -819,7 +819,8 @@ class URL(object):
         return bool(self.scheme and self.host)
 
     def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
-                fragment=_UNSET, port=_UNSET, rooted=_UNSET, userinfo=_UNSET):
+                fragment=_UNSET, port=_UNSET, rooted=_UNSET, userinfo=_UNSET,
+                family=_UNSET, uses_netloc=_UNSET):
         """:class:`URL` objects are immutable, which means that attributes
         are designed to be set only once, at construction. Instead of
         modifying an existing URL, one simply creates a copy with the
@@ -861,6 +862,8 @@ class URL(object):
             port=_optional(port, self.port),
             rooted=_optional(rooted, self.rooted),
             userinfo=_optional(userinfo, self.userinfo),
+            family=_optional(family, self.family),
+            uses_netloc=_optional(uses_netloc, self.uses_netloc)
         )
 
     @classmethod
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -141,7 +141,6 @@ _HEX_CHAR_MAP = dict([((a + b).encode('ascii'),
                       for a in string.hexdigits for b in string.hexdigits])
 _ASCII_RE = re.compile('([\x00-\x7f]+)')
 
-
 # RFC 3986 section 2.2, Reserved Characters
 #   https://tools.ietf.org/html/rfc3986#section-2.2
 _GEN_DELIMS = frozenset(u':/?#[]@')
@@ -158,6 +157,16 @@ _FRAGMENT_SAFE = _UNRESERVED_CHARS | _PATH_SAFE | set(u'/?')
 _FRAGMENT_DELIMS = _ALL_DELIMS - _FRAGMENT_SAFE
 _QUERY_SAFE = _UNRESERVED_CHARS | _FRAGMENT_SAFE - set(u'&=+')
 _QUERY_DELIMS = _ALL_DELIMS - _QUERY_SAFE
+
+
+_PATH_HEX_MAP = dict(_HEX_CHAR_MAP)
+for delim in _PATH_DELIMS:
+    _hexord = hex(ord(delim))[2:].encode('ascii')
+    try:
+        _PATH_HEX_MAP.pop(_hexord.lower())
+        _PATH_HEX_MAP.pop(_hexord.upper())
+    except KeyError:
+        print('not present:', delim)
 
 
 def _make_quote_map(safe_chars):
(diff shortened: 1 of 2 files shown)


hyperlink/test/test_url.py
```diff
@@ -115,6 +115,29 @@
...
+    'https://example.com/%3F/',  # question mark in path
+    'https://example.com/%23/',  # hash in path
+    'https://example.com/?%23=b',  # hash in query param name
+    'https://example.com/?%3D=b',  # equals in query param name
+    'https://example.com/?%26=b',  # ampersand in query param name
+    'https://example.com/?a=%23',  # hash in query param value
+    'https://example.com/?a=%26',  # ampersand in query param value
+    'https://example.com/?a=%3D',  # equals in query param value
+    # double-encoded percent sign in all percent-encodable positions:
+    "http://(%2525):(%2525)@example.com/(%2525)/?(%2525)=(%2525)#(%2525)",
```
glyph: 👍

hyperlink/test/test_url.py
```diff
@@ -231,9 +254,21 @@ def test_roundtrip(self):
         L{URL.to_text} should invert L{URL.from_text}.
         """
         for test in ROUNDTRIP_TESTS:
-            result = URL.from_text(test).to_text()
+            result = URL.from_text(test).to_text(with_password=True)
             self.assertEqual(test, result)
 
+    def test_roundtrip_double_iri(self):
```
glyph: This looks like a place where it might be interesting to drop in hypothesis (since it was in fact hypothesis that caught this bug, in txacme's test suite)

hyperlink/_url.py
```diff
@@ -861,6 +917,8 @@ def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
             port=_optional(port, self.port),
             rooted=_optional(rooted, self.rooted),
             userinfo=_optional(userinfo, self.userinfo),
+            family=_optional(family, self.family),
+            uses_netloc=_optional(uses_netloc, self.uses_netloc)
```
glyph: This seems incorrect.  You shouldn't be able to specify `uses_netloc` to the constructor at all, it's entirely a function of the scheme; adding it to `replace` seems to be propagating the error.

glyph: Confirming this, deleting these added lines doesn't make any tests fail; if it should be here at all, it shouldn't be in this change.

hyperlink/test/test_url.py
mahmoud: yeah, i'm down for some hypothesis in the future. i played with it a bit on this PR, but haven't delved deep enough yet.

hyperlink/_url.py
mahmoud: Functionally, `uses_netloc` is there because for unrecognized schemes parsed using `from_text`, URL will persist whether or not the original URL used the netloc slashes.

Logistically, I can't remember if I intended to include it in this PR or if I just saw the big regression of not passing through `family` and `uses_netloc` -- both in the constructor and the docstring of `replace`, but missing in the function signature itself -- and fixed it post haste.

glyph: If it's worth working it's worth testing :-). Urgency notwithstanding it seems like an unrelated change.

mahmoud: Well, the branch started out as seeking to normalize decoding in general, but it became pretty percent-decoding centric, as idna improvements will come in another PR. 

I've been brainstorming other ways to go with some of these latter-day arguments, but for now I think I prefer bringing things into sync, so I've added a test.

glyph:
OK, I guess the documentation already indicates that this is possible.

That said: has a release gone out with this parameter in it?  I think it's an ugly wart on the interface and I'd prefer to find a different way to deal with this if we can before committing to public API for it.

hyperlink/_url.py
```diff
@@ -160,6 +152,19 @@ def __nonzero__(self):
 _QUERY_DELIMS = _ALL_DELIMS - _QUERY_SAFE
 
 
+def _make_decode_map(delims, allow_percent=False):
+    ret = dict(_HEX_CHAR_MAP)
+    if not allow_percent:
+        delims = set(delims) | set([u'%'])
+    for delim in delims:
+        _hexord = hex(ord(delim))[2:].zfill(2).encode('ascii').upper()
```
alexwlchan: minor: I think you could simplify this by using the `%x` format tokens:

```python
_hexord = format(ord(delim), '02X').encode('ascii')
```

(I don’t know if there’s a performance difference, or if this code is particularly perf-critical)

mahmoud: This code is executed 4x at startup. Each time the loop happens <10 times. ;)

glyph: It's shorter and easier to read, so the performance issue may be secondary :)

glyph: In any case a follow-up PR to do this would be fine.

mahmoud:
@glyph yes, it's had this parameter from the initial release. It's deep enough down the parameter list that practically I don't consider it a public API. That said, with an immutable object like this, there's no other way to control whether the slashes should be emitted.

If you're wanting to slim down the parameter list, I have an idea on how to get rid of `family`... For another PR of course :)

Issue #26, opened 2017-07-03
remove "family" attribute


glyph:
I similarly ( #25 ) don't understand what this is for.  Can it just be removed?

mahmoud:
I agree this one isn't ideal, but I'm not confident my other idea for an API is more ideal, yet.

First, `family` exists to make the URL more useful for `socket` interoperation. Several `socket` APIs take a `family` argument, and this exactly maps to that. Furthermore, if `family` is `socket.AF_INET6`, then `to_text` will wrap the hostname in square brackets, per the standard. But square brackets are a no-go for socket APIs. 

In short, with the (`host`, `port`, `family`) split arrangement, `URL` is out-of-the-box usable with `socket`.

---

As for alternatives, one option I considered before hyperlink and even now, is to allow square brackets in the value for `host`, and use those as the IPv6 flag. There are a couple downsides to this:

* Users have to manually wrap/unwrap square brackets (or use a separate/new API)
* Doesn't give any indication as to whether the address parsed as IPv4

But at least it would be less likely to have family and host get out of sync. Right now,  to .replace()` a URL from an IPv4 to an IPv6 host will also require passing `family=socket.AF_INET6` (because of the colons). As I type this, I think that may be why I made `parse_host` part of the public API (#25). I think I preferred that slightly to manual IPv6 guessing and square bracket wrapping.

What do you think?

glyph:
My go-to guide to socket interoperation for non-experts is generally:

1. don't use the socket module directly, it's harder than you think
2. there is no step 2

For public-facing API, I think this is an ill-considered way to try to be "helpful".  For example, `socket.socket((url.family, socket.SOCK_STREAM))` will just raise an exception unless the URL is specified with an IP address literal, so it's not generally usable.  In the case where a hostname is specified, defaulting to either IPv4 or IPv6 is just ''wrong''; you have to do both, and it's more complicated than just issuing a single `.connect` call.  So this isn't even just me being "use twisted for your sockets"; I legitimately can't imagine what kind of application this would be helpful in.

Internally in the data representation, this causes bugs.  If someone were to read these docs and "helpfully" pass in `socket.AF_INET6` because they wanted to be sure to be IPv6 ready, you'd get this hot pile of nonsense:

```python
>>> hyperlink.URL(family=socket.AF_INET6, host=u'www.example.com')
URL.from_text(u'http://[www.example.com]')
```

[...]

PR #30, opened 2017-07-07
Defamiliarization

mahmoud:
> *Defamiliarization or ostranenie (остранение) is the artistic technique of presenting to audiences common things in an unfamiliar or strange way in order to enhance perception of the familiar.*  -- (unrelated, but relevantly named)

OK, in a classic "this, but-not-this" sort of early review step, here is a sketch of URL without `family`, per #26. I think this is probably pretty close to what we want for that issue, but there's also probably feedback, particularly from @glyph, so let's have it :)

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -136,8 +136,6 @@ _AUTHORITY_RE = re.compile(r'^(?:(?P<userinfo>[^@/?#]*)@)?'
                            r'|(?P<bad_host>.*?))?'
                            r'(?::(?P<port>.*))?$')
 
-# ^((?P<scheme>[^:/?#]+):)?((?P<_netloc_sep>//)(?P<authority>(?P<userinfo>[^@/?#]*@)?(?P<host>(?P<new_host>\[[^\[]/?#]*\])|(?P<old_host>[^\[]:/?#]*))?:?(?P<port>\d+)?))?(?P<path>[^?#]*)(\?(?P<query>[^#]*))?(#(?P<fragment>.*))?
-
 
 _HEX_CHAR_MAP = dict([((a + b).encode('ascii'),
                        unichr(int(a + b, 16)).encode('charmap'))
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -122,11 +122,30 @@ _UNRESERVED_CHARS = frozenset('~-._0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
 
 # URL parsing regex (based on RFC 3986 Appendix B, with modifications)
 _URL_RE = re.compile(r'^((?P<scheme>[^:/?#]+):)?'
-                     r'((?P<_netloc_sep>//)(?P<authority>[^/?#]*))?'
+                     r'((?P<_netloc_sep>//)'
+                     r'(?P<authority>'
+                     r'(?P<userinfo>[^@/?#]*@)?'
+                     r'(?P<host>(\[[^\[]/?#]*\])|([^:/?#[\]]*))?'
+                     r':?(?P<port>\d+)?'
+                     r'))?'
                      r'(?P<path>[^?#]*)'
                      r'(\?(?P<query>[^#]*))?'
                      r'(#(?P<fragment>.*))?')
 _SCHEME_RE = re.compile(r'^[a-zA-Z0-9+-.]*$')
+
+
+_URL_RE_PATT = (r'^((?P<scheme>[^:/?#]+):)?'
+                r'((?P<_netloc_sep>//)'
+                r'(?P<authority>'
+                r'(?P<userinfo>[^@/?#]*@)?'
+                r'(?P<host>(?P<new_host>\[[^[\]/?#]*\])|(?P<old_host>[^:/?#[\]]*))?'
+                r':?(?P<port>\d+)?'
+                r'))?'  # close authority group
+                r'(?P<path>[^?#]*)'
+                r'(\?(?P<query>[^#]*))?'
+                r'(#(?P<fragment>.*))?')
+_URL_RE = re.compile(_URL_RE_PATT)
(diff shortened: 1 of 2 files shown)


hyperlink/test/test_url.py
```diff
@@ -871,15 +871,15 @@ def test_invalid_ipv6(self):
...
-    def test_ip_family_detection(self):
-        u = URL.from_text('http://giggle.com')
-        self.assertEqual(u.family, None)
-
-        u = URL.from_text('http://127.0.0.1/a/b/?c=d')
-        self.assertEqual(u.family, socket.AF_INET)
-
-        u = URL.from_text('http://[::1]/a/b/?c=d')
-        self.assertEqual(u.family, socket.AF_INET6)
+    # def test_ip_family_detection(self):
```
glyph: https://nedbatchelder.com/text/deleting-code.html

hyperlink/test/test_url.py
```diff
@@ -856,7 +856,7 @@ def test_ipv6_with_port(self):
         url = URL.from_text(t)
         assert url.host == '2001:0db8:85a3:0000:0000:8a2e:0370:7334'
         assert url.port == 80
-        assert url.family == socket.AF_INET6
+        # assert url.family == socket.AF_INET6
```
glyph: https://nedbatchelder.com/text/deleting-code.html

hyperlink/_url.py
```diff
@@ -907,7 +888,6 @@ def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
             port=_optional(port, self.port),
             rooted=_optional(rooted, self.rooted),
             userinfo=_optional(userinfo, self.userinfo),
-            family=_optional(family, self.family),
```
glyph: I do think we probably need a deprecation warning here.  This is public, documented functionality that's disappearing, and an exception is a rude way to find out about that.

That said, this is a very easy deprecation to do, because the passed value can more or less just be ignored.

hyperlink/_url.py
```diff
@@ -970,7 +950,9 @@ def from_text(cls, text):
                             raise URLParseError('port must not be empty')
                         raise URLParseError('expected integer for port, not %r'
                                             % port_str)
-        family, host = parse_host(host)
+        # _, host = parse_host(host)
```
glyph: https://nedbatchelder.com/text/deleting-code.html

hyperlink/_url.py
```diff
@@ -970,7 +950,9 @@ def from_text(cls, text):
...
+        if host:
+            host = host.lstrip('[').rstrip(']')
```
glyph: Could this just be represented as a modification to the RE?  I would prefer to see it there than here, because here it seems like potentially invalid uses of `[` might sneak in.  (For example, a host that begins `[[[[[`).  If the `host` group only contained the appropriate text it would be cleaner to read as well.

hyperlink/_url.py
```diff
@@ -865,7 +849,7 @@ def absolute(self):
 
     def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
                 fragment=_UNSET, port=_UNSET, rooted=_UNSET, userinfo=_UNSET,
-                family=_UNSET, uses_netloc=_UNSET):
```
glyph: Another interesting concern about compatibility - someone attempting to pass either `family` *positionally* will now instead, without any error, give that value to `uses_netloc` instead, which might not have the desired effect.

hyperlink/test/test_url.py
mahmoud: 🙄

hyperlink/_url.py
mahmoud: Positional arguments are kind of a vice of the API held over for consistency. The order's kinda funky, too. I honestly believe if someone's used >4 positional arguments in an invocation, they've set themselves up the bomb.

hyperlink/_url.py
mahmoud: Have any examples handy? Just that family's not been part of the public API for very long (and indeed in the case of replace, hasn't been released at all, speaking of rude).

hyperlink/_url.py
mahmoud: Yeah I can swing that.

mahmoud: Also, it's definitely not a "just". The RE is pretty directly translated from the RFC which does not break out userinfo, host, and port, it's all "authority".

hyperlink/_url.py
glyph: If that's the case, then we need to rewrite this constructor to take `**kw` to avoid self-bombing.

hyperlink/_url.py
glyph: Heh. There should be an online swear jar for anyone who ever says "just" in a code review :)

hyperlink/_url.py
glyph: OK, good point on `replace` :)

hyperlink/_url.py
glyph: (My kingdom a kwonly arg on py2.)

hyperlink/_url.py
glyph: And, further, upon remembering that `replace` never even had it, I suppose I could be convinced that this is an outlier sufficiently extreme in its low level of usefulness and short lifetime that it's not worth the additional effort to deprecate.  (I really wish github had a code search which could help open source maintainers with this kind of thing though.)

hyperlink/_url.py
mahmoud: After about an hour and a half, I've got an RE and code that mostly does this. It _may_ be better overall, but when I push you'll see how many changes are necessary, especially to keep the error messages friendly (regexes are pretty all-or-nothing).

glyph: Comment updated.

hyperlink/_url.py
mahmoud: Yeah, if it were me starting from scratch, I'd take everything after fragment and `**kw` it, I think. In this case I'm not sure how to rewrite it, except that it's probably for another PR :)

hyperlink/_url.py
mahmoud: You can now witness the new `_AUTHORITY_RE` action, with improved "bad host" detection. Adding a few more tests before I'd consider it final. (Note that this doesn't use regex IPv6 validation, just the host syntax)

mahmoud: (tests added)

PR #33, opened 2017-07-16
adding initial MANIFEST.in

mahmoud:
Adding a MANIFEST.in to include a few files (LICENSE, README, docs). Not sure if there's a python-hyper standard approach to this, so feedback is welcome!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
       env: TOXENV=py36
     - python: "pypy"
       env: TOXENV=pypy
+    - python: "2.7"
+      env: TOXENV=packaging
 
 
 install:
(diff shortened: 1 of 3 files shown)


alexwlchan:
python-hyper “standard”, insofar as there is one, is to include everything from the repo in the sdist that might be useful – see https://github.com/python-hyper/hyper-h2/pull/501

(On a phone, so can’t double-check the proposed MANIFEST.in against the repo contents right now.)

Lukasa:
Yeah, so my strong recommendation would be to add the testing from python-hyper/hyper-h2#501 to validate that the Manifest just basically contains the entire git repo. This seems to be what's most popular with downstream repackagers.

However, if you don't want to do that, that's fine. Just let me know and I'll review with that in mind.

mahmoud:
Alright, python-hyper/hyper-h2#501 is fully implemented. Added more files, plus the `check-manifest` in a `packaging` tox env. Thanks for the advice and review!

alexwlchan:
Don’t forget to add the new test to CI.

Lukasa:
Yup, per @alexwlchan's note we should add the new test env to the CI. :grin:

PR #34, opened 2017-07-25
Add URL.normalize()

mahmoud:
This PR adds a `normalize` method to `URL`, codifying five key normalization behaviors documented in RFC 3986:

* Decode unreserved characters (`RFC 3986 2.3`)                        
* Uppercase remaining percent-encoded octets (`RFC 3986 2.1`)          
* Convert scheme and host casing to lowercase (`RFC 3986 3.2.2`)       
* Resolve any "." and ".." references in the path (`RFC 3986 6.2.2`)   
* Ensure an ending slash on URLs with an empty path (`RFC 3986 6.2.3`) 

Which looks something like:

```python
  >>> url = URL.from_text(u'Http://example.COM/a/../b/./c%2f?%61')         
  >>> url.normalize().to_text()
  u'http://example.com/b/c%2F?a'
```

This gives explicit control to users in a lot of gray areas where it wasn't clear whether we should be eager in enforcing these normal forms. Now, interested parties can explicitly normalize if they'd like. Reviews welcome!

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -6,13 +6,18 @@
 * Get coverage up
 * Switch off ctypes/socket for IP validation
 * rebase method for path (prepends to path)
+* switch default percent encoding to upper case (a la RFC 3986 2.1)
 
 ## normalize method
 
 * unquote all unreserved characters (RFC 3986 2.3)
 * lowercase scheme
-* lowercase host
+* lowercase host (including lowercase ipv6 addresses)
 * resolve path parts (RFC 3986 6.2.2)
+* percent encoding (RFC 3986 2.1) to uppercase (add a
+  `normalize_case=False` to `_percent_decode`, with appropriate upward
+  propagation)
+
 
 ## Complete
 
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1005,8 +1005,11 @@ class URL(object):
         if host:
             kw['host'] = self.host.lower()
         if path:
-            kw['path'] = [_decode_unreserved(p) for p in
-                          _resolve_dot_segments(self.path)]
+            if self.path:
+                kw['path'] = [_decode_unreserved(p) for p in
+                              _resolve_dot_segments(self.path)]
+            else:
+                kw['path'] = (u'',)
         if query:
             kw['query'] = [(_decode_unreserved(k), _decode_unreserved(v))
                            for k, v in self.query]
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -434,33 +434,45 @@ def _textcheck(name, value, delims=frozenset(), nullable=False):
     return value
 
 
-def _decode_unreserved(text):
-    return _percent_decode(text, _decode_map=_UNRESERVED_DECODE_MAP)
+def _decode_unreserved(text, normalize_case=False):
+    return _percent_decode(text, normalize_case=normalize_case,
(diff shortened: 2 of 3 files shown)


mahmoud:
One open question I still have is whether or not to do normalization on the `userinfo` or wait until we have `username`/`password` figured out. I combed through RFC3986 pretty well, didn't see any super relevant normalization there, maybe it's ok to defer on that feature.

hyperlink/_url.py
```diff
@@ -481,13 +504,26 @@ def _percent_decode(text, _decode_map=_HEX_CHAR_MAP):
...
+            try:
+                append(_decode_map[item[:2]])
+                append(item[2:])
+            except KeyError:
+                append(b'%')
+                if item[:2] in _HEX_CHAR_MAP:
+                    append(item[:2].upper())
+                    append(item[2:])
+                else:
+                    append(item)
```
mahmoud: Ah, yes, because `_percent_decode` is called so frequently, I did a bit of optimization there. Checking the branch just once has a measurable effect on longer percent-encoded strings (and I've seen some pretty big ones). I can add a comment to this effect.

glyph: Wouldn't the JIT just recognize the loop and inline the second branch automatically?

hyperlink/test/test_url.py
```diff
@@ -1100,3 +1100,37 @@ def test_from_text_bad_authority(self):
         self.assertRaises(URLParseError, URL.from_text, 'http://127.0.0.1:hi')
         # extra port colon (makes for an invalid host)
         self.assertRaises(URLParseError, URL.from_text, 'http://127.0.0.1::80')
+
+    def test_normalize(self):
+        url = URL.from_text('HTTP://Example.com/A%61/./../A%61?B%62=C%63#D%64')
+        assert url.get('Bb') == []
```
glyph: This should be using `assertEquals`; the `assert` keyword will be discarded when run under `python -OO`, making it impossible to test this in a memory-constrained environment.

TODO.md
```diff
@@ -6,13 +6,18 @@
 * Get coverage up
 * Switch off ctypes/socket for IP validation
 * rebase method for path (prepends to path)
+* switch default percent encoding to upper case (a la RFC 3986 2.1)
 
 ## normalize method
```
glyph: Should this section be deleted?

hyperlink/_url.py
```diff
@@ -982,6 +1018,60 @@ def from_text(cls, text):
...
+        http://example.com/b/c%2F?a
+
+        .. _RFC 3986 3.2.2: https://tools.ietf.org/html/rfc3986#section-3.2.2
+        .. _RFC 3986 2.3: https://tools.ietf.org/html/rfc3986#section-2.3
+        .. _RFC 3986 2.1: https://tools.ietf.org/html/rfc3986#section-2.1
+        .. _RFC 3986 6.2.2.3: https://tools.ietf.org/html/rfc3986#section-6.2.2.3
+        .. _RFC 3986 6.2.3: https://tools.ietf.org/html/rfc3986#section-6.2.3
+
+        """
+        # TODO: userinfo?
```
glyph: This should reference a URL to a ticket or something, or possibly just be deleted (otherwise it will persist forever, likely even after userinfo normalization has been implemented :))

Issue #35, opened 2017-08-04
Add an API for resolving a URL against another


Julian:
Hi!

`url.click` appears to take a str. It seems like it would be nice to have an API that takes a `URL` instead.

(For the same reasons why dealing with `URL`s and not `str`s is better in the first place :)

mahmoud:
You got it! I made the fix in #36. If you wouldn't mind taking a look, I'd appreciate it a lot. If it looks good to you, give me an LGTM comment, and I'll merge it :)

Julian:
Awesome! Thanks for the quick turnaround!

PR #36, opened 2017-08-05
URL.click() for URL objects

mahmoud:
This PR makes URL.click work with URL objects, not just text, fixing #35.

There is one slightly outstanding issue, detailed in a `TODO` comment, which is that I need to remove the error message checks in Twisted's test suite. I haven't known anybody to depend on the byte-for-byte unchangingness of error messages as a public API, and in this case, it's restricting usability.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1049,9 +1049,15 @@ class URL(object):
 
         .. _RFC 3986 section 5: https://tools.ietf.org/html/rfc3986#section-5
         """
-        _textcheck("relative URL", href)
         if href:
-            clicked = URL.from_text(href)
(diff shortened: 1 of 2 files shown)


hyperlink/_url.py
```diff
@@ -1049,9 +1049,15 @@ def click(self, href=u''):
...
         """
-        _textcheck("relative URL", href)
         if href:
-            clicked = URL.from_text(href)
+            if isinstance(href, URL):
+                clicked = href
+            else:
+                # TODO: This error message is not completely accurate,
+                # as URL objects are now also valid, but Twisted's
+                # test suite (wrongly) relies on this exact message.
```
Julian: Agreed

Issue #38, opened 2017-09-06
Excessive escaping of "=" in query string parameter values


mahmoud:
As a more specific continuation of the discussion in #11, it would seem that the `=` character is yet another special case. While `=` is a meaningful character in the query string, separating keys and values, only the first `=` does that.

Digging in further, empty query parameter keys are OK. And equals signs in the value of query parameter values are OK.

```
# Werkzeug request object for "GET http://localhost:5000/?=x=x=x="
# from Firefox and Chrome
(Pdb) request.args
ImmutableMultiDict([('', u'x=x=x=')])
```

Seen here, and in their developer tools, Firefox and Chrome do not encode the equals signs. On the server side, Werkzeug is ok with this.

Now, urllib does, but I think this is only because their implementation is lazy. :)

mahmoud:
The example above isn't from nowhere, it's [from `twisted.python.test.test_url`](https://github.com/twisted/twisted/blob/02a3221ed72e3385546f9421401bf40e360bb9bb/src/twisted/python/test/test_url.py#L484). The problem I see is that Twisted expects urllib's overzealous escaping behavior: 

```python
self.assertEqual(u.asText(), 'http://localhost/?=x%3Dx%3Dx')
```

And I think that should change to simply roundtrip to `'http://localhost/?=x=x=x'`. Thoughts, @glyph?

glyph:
Given that =x=x=x is indeed parseable, let's follow our usual convention here and leave it as-is, as you suggest.

When we address #11, we may want to be a little more zealous, given that the closest reading of https://url.spec.whatwg.org/#concept-urlsearchparams-update that I have the time for suggests (via https://url.spec.whatwg.org/#concept-urlencoded-byte-serializer) that = *should* be %-encoded when updating query params.

mahmoud:
Cool. I've created [#9257 in trac](https://twistedmatrix.com/trac/ticket/9257) for when I actually get around to loosening up those tests in Twisted.

As for the WHATWG, I think they're trying to flex muscle they doesn't have. If Chrome and Firefox don't do it, I'm not sure which Hypertext Applications do, you know? For a variety of reasons I can go into if you'd like, my preference is RFC3986 (and certain other non-obsolete RFCs) followed by browser behavior and then, somewhere down the list, WHATWG.

glyph:
[...]

PR #39, opened 2017-09-06
Allow unescaped equal signs in query parameter values

mahmoud:
Specifically, allow unescaped equal signs in query parameter values. Values are terminated by ampersands (`&`) or the fragment (`#`) or the end of the URL. See #38 for more details.

This change requires changing [Twisted's tests](https://github.com/twisted/twisted/blob/02a3221ed72e3385546f9421401bf40e360bb9bb/src/twisted/python/test/test_url.py#L484-L490), which inherited the overzealous escaping of urllib.

### DIFF
--- a/hyperlink/test/test_url.py
+++ b/hyperlink/test/test_url.py
@@ -532,10 +532,16 @@ class TestURL(HyperlinkTestCase):
         """
         u = URL.from_text('http://localhost/?=x=x=x')
         self.assertEqual(u.get(''), ['x=x=x'])
-        self.assertEqual(u.to_text(), 'http://localhost/?=x%3Dx%3Dx')
+        # TODO: see #38
+        # self.assertEqual(u.to_text(), 'http://localhost/?=x%3Dx%3Dx')
         u = URL.from_text('http://localhost/?foo=x=x=x&bar=y')
         self.assertEqual(u.query, (('foo', 'x=x=x'), ('bar', 'y')))
-        self.assertEqual(u.to_text(), 'http://localhost/?foo=x%3Dx%3Dx&bar=y')
+        # TODO: see #38
+        # self.assertEqual(u.to_text(), 'http://localhost/?foo=x%3Dx%3Dx&bar=y')
+
+        u = URL.from_text('https://example.com/?argument=3&argument=4&operator=%3D')
+        iri = u.to_iri()
+        self.assertEqual(iri.get('operator'), ['='])
 
     def test_empty(self):
         """
--- a/hyperlink/test/test_url.py
+++ b/hyperlink/test/test_url.py
@@ -542,12 +542,10 @@ class TestURL(HyperlinkTestCase):
         """
         u = URL.from_text('http://localhost/?=x=x=x')
         self.assertEqual(u.get(''), ['x=x=x'])
-        # TODO: see #38
-        # self.assertEqual(u.to_text(), 'http://localhost/?=x%3Dx%3Dx')
+        self.assertEqual(u.to_text(), 'http://localhost/?=x=x=x')
         u = URL.from_text('http://localhost/?foo=x=x=x&bar=y')
         self.assertEqual(u.query, (('foo', 'x=x=x'), ('bar', 'y')))
-        # TODO: see #38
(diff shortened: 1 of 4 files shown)


hyperlink/test/test_url.py
```diff
@@ -532,10 +532,18 @@ def test_parseEqualSignInParamValue(self):
         """
         u = URL.from_text('http://localhost/?=x=x=x')
         self.assertEqual(u.get(''), ['x=x=x'])
-        self.assertEqual(u.to_text(), 'http://localhost/?=x%3Dx%3Dx')
+        # TODO: see #38
```
glyph: Why is this a TODO?  Isn't #38 the bug that specifically asks this *not* to be done?

glyph:
@mahmoud Are you planning to move this along so it can land?

glyph:
Time for my monthly bump here :)

glyph:
I can't find the original citation, but I believe I had encountered a case where I built a URL with multiple =-signs in it for feeding to an API, and the API did not work with the escaped version.

Per WHATWG, browsers (tested: Safari, Chrome, Firefox) definitely will not quote = signs in this position if they're present in the address bar.  So I think this is the correct thing to do, interop-wise.

mahmoud:
I guess as long as web servers are ok with it, too, that's somewhat of a quorum. Kind of weird they wouldn't have a more general decoding policy; unreserved characters get encoded all the time.  Speaking of time, I still don't have the time to make the Twisted change happen, but if someone wants to yank out some of those tests, I'll meet you halfway and fix merge conflicts, etc. get this mergeable again.

glyph:
@twisted/twisted-contributors anybody want to take a look at removing this overzealous test for library code from the Twisted suite?

glyph:
Erm. How does @-mentioning teams work

glyph:
https://blog.github.com/2012-05-09-introducing-team-mentions/ suggests that this should have worked?

glyph:
> Kind of weird they wouldn't have a more general decoding policy; unreserved characters get encoded all the time

Oh, for sure.  If the server behaves correctly, either representation should be OK.  But this is exactly why hyperlink has to allow for multiple possible quoting representations ;).

glyph:
Bump…

glyph:
👉

glyph:
⁉️

glyph:
Hey @mahmoud - any thoughts on fixing this up?

glyph:
🎶

mahmoud:
lol ok today's your lucky day, I'm puttin everything else on hold takin a look! (for however long this train ride takes :P )

hyperlink/test/test_url.py
mahmoud: test updated. basically this was a legacy twisted test that I was worried about breaking. The TODO was more like "TODO: fix this in Twisted's test suite". Does Twisted even still have the t.p.url tests duplicated in it? If so those will have to change/come out before updating to the 19.x hyperlink release. :)

mahmoud: Yeah doesn't look like the Twisted test suite has had the deep tests removed yet: https://github.com/twisted/twisted/blob/8292869715d9471681ffff32296bc73b946868a3/src/twisted/python/test/test_url.py#L479

Not sure why that should block a hyperlink review/release though. Whomever does the upgrade on the Twisted side (i.e., me or you, but maybe someone else) can remove the corresponding tests based on info that'll be in the hyperlink changelog.

glyph: In general I feel like one should feel badly about `pip install` of the latest version causing downstreams' tests to fail with ~no warning ;).  That said, this might be what Twisted's own compatibility policy might call a "gross violation of specifications" and there's no compatible fix-forward way to handle this (i.e.: twisted is just directly testing the wrong thing); so I remove the objection unless you can think of a way to emit a warning rather than fail.

glyph:
Hooray, the system works

glyph:
@mahmoud 

> Note: feel free to review, but do not merge.

I am holding off on merging myself because this comment is still in the PR description, but if your views have moved on from this perspective, please feel free to land yourself.

PR #46, opened 2017-11-15
mention py36 testing in docs, not just in readme

wbolster:


mahmoud:
Oh, nice! Thanks, Wouter!

wbolster:
actually tox also runs py26, but i will never contribute anything that might lead others to think py26 is an acceptable platform. not in 2017. ;)

PR #52, opened 2017-12-05
Use an explicit check for Python 2

wbolster:
...instead of relying on a cryptic `if bytes is str` which just causes
confusion.


wbolster:
i proudly live in a python 3 only world (at least that's what i tell myself), so stuff like "str is bytes" look impossible to me... until i force myself to relive these bad childhood memories about python 2 text handling. an explicit py2 warning in the code allow me to quickly read past those ghosts from the past. ;)

PR #53, opened 2017-12-05
Also accept dictionaries as ‘query=’ arguments

### LINKED ISSUE #50: use a dictionary for query params
`hyperlink` does not support dictionaries for `query`. while i know that technically this is an `orderedmultidict`, this is just not nice from an api perspective:

``` python
>>> url = URL.from_text('https://example.org/api/v2')

>>> url.replace(query={'foo': 'bar'})
[...]
ValueError: too many values to unpack (expected 2)
```

instead, this works:

``` python
>>> url.replace(query=[('foo', 'bar')])
URL.from_text('https://example.org/api/v2?foo=bar')
```

or using `.items()`:

```python
>>> url.replace(query={'foo': 'bar'}.items())
URL.from_text('https://example.org/api/v2?foo=bar')
```

but tbh none of this looks appealing to me. :)

any reason for the current behaviour?

wbolster:
The URL constructor and the `.replace()` method now sensibly handle
dictionaries passed to the ‘query’ argument, instead of strictly
requiring it to be an iterable of `(key, value)` pairs (both of which
must be strings).

Examples:

```
  URL(query={"foo": "bar"})
  url.replace(query={"foo": "bar"})
```

This simplifies the API for the common use case, in which:

- the order of the items does not matter
  (though Python >= 3.6 preserves dictionary order)
- no duplicate parameters are present

Updated the docs to include an example of this.

Fixes #50.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## dev (not yet released)
 
+* Also accept dictionaries as ‘query=’ arguments (see #50)
+
 ## 17.3.1
 
 *(August 19, 2017)*
--- a/docs/design.rst
+++ b/docs/design.rst
@@ -92,6 +92,14 @@ by the query parameters, sometimes called "query arguments" or "GET
 parameters". Regardless of what you call them, they are encoded in
 the query string portion of the URL, and they are very powerful.
 
+In the simplest case, these query parameters can be provided as a
+dictionary:
+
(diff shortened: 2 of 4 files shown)


wbolster:
you're welcome!

not sure what you're referring to re. the docstring, but now that i look at the docstring lines touched in this pr, the "tuple" type definition was never correct since other types worked as well, and now that's even more flexible with transparent dict handling.

Issue #44, opened 2017-10-26
Recommended practice for adding reserved characters?


mahmoud:
Per @markrwilliams comment [here](https://github.com/python-hyper/hyperlink/issues/6#issuecomment-338984058) and a few others dotted around, we're facing a design gap in hyperlink's APIs. 

To paraphrase:
```
url = URL()
url.add(u'param', u'#value').to_text()
```
Yields:
```
ValueError: one or more reserved delimiters &# present in query parameter value: u'#value'
```

This is due to a subtle shift in hyperlink's design compared to `twisted.python.url`. `t.p.url` would allow any string value in, whereas hyperlink prefers to store the "minimally-encoded" version. This is why a `ValueError` is raised from the code above.

Technically, this can be solved by making the code `url.add(u'param', _encode_query_part(u'#value'))`. But hyperlink's primary goal is to handle encoding/decoding, does it really make sense to push that back on the user?

One solution Mark and I discussed would be to switch to decoding every value passed in. But what if someone were to pass in `u'%23%'` and actually intend for that to be their decoded value? And the API would be further complicated by the fact that the underlying decoding is generally unknown. UTF8, Latin-1, and plain old binary are all valid in percent-encoded URL parts. Autodecoding UTF8 might have better usability most of the time, but much like relying on Python 2's implicit encoding/decoding, the safety of the explicit `_encode_*_part()` is probably preferable.

It might occur to one that this entire problem bears some resemblance to the `bytes`/`unicode` split, as `URL` has `URL.to_uri()` and `URL.to_iri()`. There is some truth to this, but both IRIs and URIs are both URLs. Having two types imposes a sort of artificial split I'd like to avoid if possible, but we also don't have a good way to represent an already decoded IRI. This was causing an issue with double decoding on multiple `.to_iri()` calls (see #16). 

Right now my best idea is to enable that technical solution above by exposing the various encoding and decoding functions as public APIs, since those may prove useful utilities for other contexts anyways. I'm sure there are better ideas, too, so I'm going to leave this issue open as a place for discussion on handling this quandary.

glyph:
IRIs vs. URLs is a red herring here.  *reserved* characters in a URL are still reserved in an IRI; the decoding that happens is all reversible and idempotent.

[...]

PR #54, opened 2017-12-31
DecodedURL

mahmoud:
This still needs some documentation, but to address #44 and the handful of issues surrounding the central problem, I present `DecodedURL`. It takes care of handling reserved characters so you don't have to. 

From the docstring:

> DecodedURL is a type meant to act as a higher-level interface to the URL. It is the unicode to URL's bytes. DecodedURL has almost exactly the same API as URL, but everything going in and out is in its maximally decoded state. All percent decoding is handled automatically.
> 
> Where applicable, a UTF-8 encoding is presumed. Be advised that some interactions, can raise UnicodeEncodeErrors and UnicodeDecodeErrors, just like when working with bytestrings.
> 
> Examples of such interactions include handling query strings encoding binary data, and paths containing segments with special characters encoded with codecs other than UTF-8. 

It's tested, works, and seems practical, though, so take a look!

### DIFF
--- a/hyperlink/__init__.py
+++ b/hyperlink/__init__.py
@@ -1,10 +1,16 @@
 
-from ._url import URL, DecodedURL, URLParseError, register_scheme, parse_host
+from ._url import (URL,
+                   parse,
+                   EncodedURL,
+                   DecodedURL,
+                   URLParseError,
+                   register_scheme)
 
 __all__ = [
     "URL",
+    "parse",
+    "EncodedURL",
     "DecodedURL",
     "URLParseError",
     "register_scheme",
-    "parse_host"
 ]
--- a/hyperlink/__init__.py
+++ b/hyperlink/__init__.py
@@ -1,8 +1,9 @@
 
-from ._url import URL, URLParseError, register_scheme, parse_host
+from ._url import URL, DecodedURL, URLParseError, register_scheme, parse_host
 
 __all__ = [
     "URL",
+    "DecodedURL",
     "URLParseError",
     "register_scheme",
     "parse_host"
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1468,6 +1468,11 @@ class DecodedURL(object):
                       for x in (k, v))
                 for k, v in self._url.query]
 
+    @property
+    def userinfo(self):
+        return u':'.join([_percent_decode(p) for p in
+                          self._url.userinfo.split(':', 1)])
+
     def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
                 fragment=_UNSET, port=_UNSET, rooted=_UNSET, userinfo=_UNSET,
                 uses_netloc=_UNSET):
@@ -1480,7 +1485,7 @@ class DecodedURL(object):
                      for k, v in iter_pairs(query)]
         if userinfo is not _UNSET:
             userinfo = u':'.join([_encode_userinfo_part(p) for p in
-                                  self.userinfo.split(':', 1)])
+                                  userinfo.split(':', 1)])
         new_url = self._url.replace(scheme=scheme,
                                     host=host,
                                     path=path,
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1519,9 +1519,9 @@ class DecodedURL(object):
         return
 
     @classmethod
-    def from_text(cls, text):
+    def from_text(cls, text, lazy=False):
         _url = URL.from_text(text)
-        return cls(_url)
+        return cls(_url, lazy=lazy)
 
     @property
     def encoded_url(self):
@@ -1623,13 +1623,6 @@ class DecodedURL(object):
         return self.userinfo[0]
 
     @property
-    def password(self):
-        try:
-            return self.userinfo[1]
-        except IndexError:
-            return None
-
-    @property
     def uses_netloc(self):
         return self._url.uses_netloc
 
@@ -1711,8 +1704,8 @@ class DecodedURL(object):
     asIRI = to_iri
 
     @classmethod
-    def fromText(cls, s):
-        return cls.from_text(s)
+    def fromText(cls, s, lazy=False):
+        return cls.from_text(s, lazy=lazy)
 
     def asText(self, includeSecrets=False):
         return self.to_text(with_password=includeSecrets)
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -202,11 +202,25 @@ _QUERY_PART_QUOTE_MAP = _make_quote_map(_QUERY_SAFE)
 _QUERY_DECODE_MAP = _make_decode_map(_QUERY_DELIMS)
 _FRAGMENT_QUOTE_MAP = _make_quote_map(_FRAGMENT_SAFE)
 _FRAGMENT_DECODE_MAP = _make_decode_map(_FRAGMENT_DELIMS)
+_UNRESERVED_QUOTE_MAP = _make_quote_map(_UNRESERVED_CHARS)
 _UNRESERVED_DECODE_MAP = dict([(k, v) for k, v in _HEX_CHAR_MAP.items()
                                if v.decode('ascii', 'replace')
                                in _UNRESERVED_CHARS])
 
 _ROOT_PATHS = frozenset(((), (u'',)))
+
+
+def _encode_reserved(text, maximal=True):
+    """A very comprehensive percent encoding for encoding all
+    delimeters. Used for arguments to DecodedURL, where a % means a
+    percent sign, and not the character used by URLs for escaping
+    bytes.
+    """
+    if maximal:
+        bytestr = normalize('NFC', text).encode('utf8')
+        return u''.join([_UNRESERVED_QUOTE_MAP[b] for b in bytestr])
+    return u''.join([_UNRESERVED_QUOTE_MAP[t] if t in _UNRESERVED_CHARS
+                     else t for t in text])
 
 
 def _encode_path_part(text, maximal=True):
@@ -1556,10 +1570,11 @@ class DecodedURL(object):
     def child(self, *segments):
         if not segments:
             return self
-        return type(self)(self._url.child(*segments))
+        new_segs = [_encode_reserved(s) for s in segments]
+        return type(self)(self._url.child(*new_segs))
 
     def sibling(self, segment):
-        return type(self)(self._url.sibling(segment))
+        return type(self)(self._url.sibling(_encode_reserved(segment)))
 
     def click(self, href=u''):
         return type(self)(self._url.click(href=href))
@@ -1603,14 +1618,13 @@ raise exceptions, or at least cachedproperties.
   * replace
   * add()
   * set()
-  * get()
+  * child (split and encode?)
+  * sibling
   * remove()
 * Passthrough
   * __eq__ / __ne__ / __hash__
   * absolute()
 * Return new DecodedURL with new ._url (the other kind of passthrough)
-  * child (split and encode?)
-  * sibling
   * normalize()
   * click()
 * Strict passthrough (doesn't return a DecodedURL)
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1506,8 +1506,16 @@ class DecodedURL(object):
 
     @property
     def userinfo(self):
-        return u':'.join([_percent_decode(p, raise_subencoding_exc=True)
-                          for p in self._url.userinfo.split(':', 1)])
+        return tuple([_percent_decode(p, raise_subencoding_exc=True)
+                      for p in self._url.userinfo.split(':', 1)])
+
+    @property
+    def user(self):
+        return self.userinfo[0]
+
+    @property
+    def password(self):
+        return self.userinfo[1]
 
     @property
     def fragment(self):
@@ -1545,6 +1553,39 @@ class DecodedURL(object):
                          for (k, v) in self._url.query]
         return [v for (k, v) in decoded_query if name == k]
 
+    def child(self, *segments):
+        if not segments:
+            return self
+        return type(self)(self._url.child(*segments))
+
+    def sibling(self, segment):
+        return type(self)(self._url.sibling(segment))
+
+    def click(self, href=u''):
+        return type(self)(self._url.click(href=href))
+
+    def normalize(self, *a, **kw):
+        return type(self)(self._url.normalize(*a, **kw))
+
+    def to_uri(self, *a, **kw):
+        return self._url.to_uri(*a, **kw)
+
+    def to_iri(self, *a, **kw):
+        return self._url.to_iri(*a, **kw)
+
+    def to_text(self, *a, **kw):
+        return self._url.to_text(*a, **kw)
+
+    def __repr__(self):
+        cn = self.__class__.__name__
+        return '%s(url=%r)' % (cn, self._url)
+
+    def __str__(self):
+        # TODO: the underlying URL's __str__ needs to change to make
+        # this work as the URL
+        return str(self._url)
+
+
 
 """Probably turn the properties into normal attributes now that they
 raise exceptions, or at least cachedproperties.
@@ -1560,8 +1601,6 @@ raise exceptions, or at least cachedproperties.
   * .get()
 * Wrap in encoder
   * replace
-  * child (split and encode?)
-  * sibling
   * add()
   * set()
   * get()
@@ -1570,6 +1609,8 @@ raise exceptions, or at least cachedproperties.
   * __eq__ / __ne__ / __hash__
   * absolute()
 * Return new DecodedURL with new ._url (the other kind of passthrough)
+  * child (split and encode?)
+  * sibling
   * normalize()
   * click()
 * Strict passthrough (doesn't return a DecodedURL)
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -485,7 +485,8 @@ def _decode_fragment_part(text, normalize_case=False):
                            _decode_map=_FRAGMENT_DECODE_MAP)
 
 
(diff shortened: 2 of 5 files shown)


hyperlink/_url.py
```diff
@@ -202,13 +202,27 @@ def _make_quote_map(safe_chars):
...
 _UNRESERVED_DECODE_MAP = dict([(k, v) for k, v in _HEX_CHAR_MAP.items()
                                if v.decode('ascii', 'replace')
                                in _UNRESERVED_CHARS])
 
 _ROOT_PATHS = frozenset(((), (u'',)))
 
 
+def _encode_reserved(text, maximal=True):
+    """A very comprehensive percent encoding for encoding all
+    delimeters. Used for arguments to DecodedURL, where a % means a
```
alexwlchan: nit: "delim*i*ters"

hyperlink/_url.py
```diff
@@ -1449,3 +1473,272 @@ def remove(self, name):
...
+            return NotImplemented
+        return self.normalize().to_uri() == other.normalize().to_uri()
+
+    def __ne__(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash((self.__class__, self.scheme, self.userinfo, self.host,
```
alexwlchan: minor: Since equality is delegated upwards to `URL`, would it be better to do the same for hashing?

e.g.

```python
def __hash__(self):
    return hash(self.normalize().to_uri())
```

hyperlink/__init__.py
```diff
@@ -1,8 +1,9 @@
 
-from ._url import URL, URLParseError, register_scheme, parse_host
+from ._url import URL, DecodedURL, URLParseError, register_scheme, parse_host
 
 __all__ = [
     "URL",
+    "DecodedURL",
```
glyph: I want to say that we should just avoid exposing this entirely, but it probably needs to be exposed for type annotations.

However, as per #44, could we have a "decoded" property on `URL` that provides an interface to this, and an "encoded" property on `DecodedURL` that maps back to a `URL`?

(At this point I think I'm in favor of adding an `EncodedURL` alias for `URL`, then maybe adding a top-level entry point like `hyperlink.parse()` which takes a `decoded` kwarg flag which defaults to true, to make it easier to get started with `DecodedURL` which is what I think we all want most of the time.  That can definitely be deferred to a separate ticket though.)

hyperlink/_url.py
mahmoud: That's a good point. I may have subtly changed my mind midway through development, and now I lean more toward not delegating up to URL for equality. I'll get that fixed, thanks!

hyperlink/__init__.py
mahmoud: Yes, I am pretty much in favor of all those conveniences :) And I also agree that this probably needs to be exposed.

hyperlink/_url.py
mahmoud: Oooh, this kind of typo is so very unlike me, I assure you! ;) Thanks!

hyperlink/_url.py
glyph: Whenever you're delegating `hash` like this, remember you should *also* include a tweak, so that the `DecodedURL` and the `URL` representing the same data don't hash the same.

moshez: @glyph why is that? (correctness shouldn't be effected since dictionaries compare by equality, and I'm not entirely sure what the performance problem you're trying to forestall)

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+    Where applicable, a UTF-8 encoding is presumed. Be advised that
+    some interactions, can raise UnicodeEncodeErrors and
+    UnicodeDecodeErrors, just like when working with
+    bytestrings.
+
+    Examples of such interactions include handling query strings
+    encoding binary data, and paths containing segments with special
+    characters encoded with codecs other than UTF-8.
+    """
+    def __init__(self, url, lazy=False):
```
markrwilliams: If this is a public class, then its initializer should be documented.  Please include an explanation of `url` and `lazy` in the docstring.

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        "Passthrough to :meth:`~hyperlink.URL.to_uri()`"
+        return self._url.to_uri(*a, **kw)
+
+    def to_iri(self, *a, **kw):
+        "Passthrough to :meth:`~hyperlink.URL.to_iri()`"
+        return self._url.to_iri(*a, **kw)
+
+    def click(self, href=u''):
+        "Return a new DecodedURL wrapping the result of :meth:`~hyperlink.URL.click()`"
+        return type(self)(self._url.click(href=href))
```
markrwilliams: Why not `self.__class__`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+    def to_uri(self, *a, **kw):
+        "Passthrough to :meth:`~hyperlink.URL.to_uri()`"
+        return self._url.to_uri(*a, **kw)
+
+    def to_iri(self, *a, **kw):
+        "Passthrough to :meth:`~hyperlink.URL.to_iri()`"
+        return self._url.to_iri(*a, **kw)
+
+    def click(self, href=u''):
+        "Return a new DecodedURL wrapping the result of :meth:`~hyperlink.URL.click()`"
```
markrwilliams: Backticks around `DecodedURL`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+
+    @property
+    def query(self):
+        try:
+            return self._query
+        except AttributeError:
+            pass
+        _q = [tuple(_percent_decode(_encode_query_part(x),
+                                    raise_subencoding_exc=True)
+                    if x is not None else None
```
markrwilliams: Can I have a key of `None` with a non-`None` value?  The `_textcheck` for the query parameter keys in `URL.__init__` isn't nullable, so I think the answer is never.  A defensive change would only check the value for `None`.

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+
+    @property
+    def uses_netloc(self):
+        return self._url.uses_netloc
+
+    def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
+                fragment=_UNSET, port=_UNSET, rooted=_UNSET, userinfo=_UNSET,
+                uses_netloc=_UNSET):
+        """While the signature is the same, this `replace()` differs a little from
+        URL.replace. For instance, it accepts userinfo as a tuple, not
```
markrwilliams: Why is it different?  Presumably it's to make the encoding easier, if not just possible, but it'd be good to see the reason discussed in the docstring.

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        as a string. As with the rest of the methods on DecodedURL, if
+        you pass a reserved character, it will be automatically
+        encoded instead of an error being raised.
+        """
+        if path is not _UNSET:
+            path = [_encode_reserved(p) for p in path]
+        if query is not _UNSET:
+            query = [[_encode_reserved(x)
+                      if x is not None else None
+                      for x in (k, v)]
```
markrwilliams: See above about nullable query parameter keys.

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+                                    query=query,
+                                    fragment=fragment,
+                                    port=port,
+                                    rooted=rooted,
+                                    userinfo=userinfo,
+                                    uses_netloc=uses_netloc)
+        return type(self)(url=new_url)
+
+    def get(self, name):
+        "Get the value of all query parameters whose name matches *name*"
```
markrwilliams: Why not copy the docstring from `URL`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        return [v for (k, v) in self.query if name == k]
+
+    def add(self, name, value=None):
+        "Return a new DecodedURL with the query parameter *name* and *value* added."
```
markrwilliams: Why not copy the docstring from `URL`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        return self.replace(query=self.query + ((name, value),))
+
+    def set(self, name, value=None):
+        "Return a new DecodedURL with query parameter *name* set to *value*"
```
markrwilliams: Why not copy the docstring from `URL`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        query = self.query
+        q = [(k, v) for (k, v) in query if k != name]
+        idx = next((i for (i, (k, v)) in enumerate(query) if k == name), -1)
+        q[idx:idx] = [(name, value)]
+        return self.replace(query=q)
+
+    def remove(self, name):
+        "Return a new DecodedURL with query parameter *name* removed."
```
markrwilliams: Why not copy the docstring from `URL`?

hyperlink/_url.py
markrwilliams: I share @moshez's curiosity.

@mahmoud - will you follow @alexwlchan's advice about delegating `__hash__`?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+    def __hash__(self):
+        return hash((self.__class__, self.scheme, self.userinfo, self.host,
+                     self.path, self.query, self.fragment, self.port,
+                     self.rooted, self.uses_netloc))
+
+    # # Begin Twisted Compat Code
+    asURI = to_uri
+    asIRI = to_iri
+
+    @classmethod
```
markrwilliams: Why is this its own `classmethod` when the instance methods are aliased?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+    def fromText(cls, s, lazy=False):
+        return cls.from_text(s, lazy=lazy)
+
+    def asText(self, includeSecrets=False):
```
markrwilliams: As above.

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+        except AttributeError:
+            # object.__dir__ == AttributeError  # pdw for py2
+            ret = dir(self.__class__) + list(self.__dict__.keys())
+        ret = sorted(set(ret) - set(['fromText', 'asURI', 'asIRI', 'asText']))
+        return ret
+
+    # # End Twisted Compat Code
+
+
+def parse(url, decoded=True, lazy=False):
```
markrwilliams: No docstring?

hyperlink/_url.py
```diff
@@ -1464,3 +1504,329 @@ def remove(self, name):
...
+    enc_url = EncodedURL.from_text(url)
+    if not decoded:
+        return enc_url
+    dec_url = DecodedURL(enc_url, lazy=lazy)
+    return dec_url
+
+
+"""Probably turn the properties into normal attributes now that they
```
markrwilliams: Looks like most of this is done.  Please remove everything except what's left to do for posterity's sake.

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
```
markrwilliams: I am very anti `unicode_literals` because of the trouble it causes with filenames, for example.  I don't expect you to remove it but I did want to mention it!

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+        assert durl.host == 'bücher.ch'
+        assert durl.port == 8080
+        assert durl.path == ('a', 'nice nice', '.', 'path', '')
+        assert durl.fragment == 'frég'
+        assert durl.get('zot') == ['23%']
+
+        assert durl.user == 'user'
+        assert durl.userinfo == ('user', '\0\0\0\0')
+
+    def test_passthroughs(self):
```
markrwilliams: A lot of these (e.g., `sibling`, `__eq__`) aren't passthroughs.  A better name for it would be `test_interface_equivalent_to_URL`.

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+
+        assert durl.absolute
+        assert durl.rooted
+
+        assert durl == durl.encoded_url.get_decoded_url()
+
+        durl2 = DecodedURL.from_text(TOTAL_URL, lazy=True)
+        assert durl2 == durl2.encoded_url.get_decoded_url(lazy=True)
+
+        # TODO change this to the actual value when str() changes get merged
```
markrwilliams: This TODO can be addressed.

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+        assert '%20' in str(DecodedURL.from_text(BASIC_URL).child(' '))
+
+        assert not (durl == 1)
+        assert durl != 1
+
+    def test_repr(self):
+        durl = DecodedURL.from_text(TOTAL_URL)
+        assert repr(durl) == 'DecodedURL(url=' + repr(durl._url) + ')'
```
markrwilliams: A soft 👎 for using a private API in the test.  Not necessary to address in this PR, but I think a canned repr would be fine given most of the tests have hard dependencies on the content of `TOTAL_URL` anyway.

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+    def test_query_manipulation(self):
+        durl = DecodedURL.from_text(TOTAL_URL)
+
+        assert durl.get('zot') == ['23%']
+        durl = durl.add(' ', 'space')
+        assert durl.get(' ') == ['space']
+        durl = durl.set(' ', 'spa%ed')
+        assert durl.get(' ') == ['spa%ed']
+
+        durl = DecodedURL(url=durl.to_uri())
```
markrwilliams: What's the point of this?  Round tripping?

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+        assert durl.get(' ') == ['spa%ed']
+        durl = durl.remove(' ')
+        assert durl.get(' ') == []
+
+        durl = DecodedURL.from_text('/?%61rg=b&arg=c')
+        assert durl.get('arg') == ['b', 'c']
+
+        assert durl.set('arg', 'd').get('arg') == ['d']
+
+    def test_equivalences(self):
```
markrwilliams: This really tests `__eq__` and `__hash__`, so maybe `test_equality_and_hashability`?

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+
+        assert durl == durl
+        assert durl == durl2
+        assert durl != burl
+        assert durl != None
+        assert durl != durl._url
+
+        durl_map = {}
+        durl_map[durl] = durl
+        durl_map[durl2] = durl2
```
markrwilliams: This should also test that a `URL` and a `DecodedURL` that represent the same underlying URL don't overlap in a `dict` (or `set`.)

hyperlink/test/test_decoded_url.py
```diff
@@ -0,0 +1,139 @@
...
+
+        assert len(durl_map) == 1
+
+    def test_replace_roundtrip(self):
```
markrwilliams: This is good, but there should also be a roundtrip test between `TOTAL_URL` and `DecodedURL`.

hyperlink/test/test_parse.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals
+
+from .common import HyperlinkTestCase
+from hyperlink import parse, EncodedURL, DecodedURL
+
+BASIC_URL = 'http://example.com/#'
+TOTAL_URL = "https://%75%73%65%72:%00%00%00%00@xn--bcher-kva.ch:8080/a/nice%20nice/./path/?zot=23%25&zut#frég"
+UNDECODABLE_FRAG_URL = TOTAL_URL + '%C3'
```
markrwilliams: It'd be good to note that this is undecodable because `%C3` makes it invalid UTF-8.

hyperlink/_url.py
mahmoud: Because what if you have a username with a colon in it, same as everything else in DecodedURL. Updating the docstring.

hyperlink/_url.py
mahmoud: started with that, but most docstrings mention URL by name.

hyperlink/_url.py
mahmoud: mentions URL by name (and has doctests which would be confusing).

hyperlink/_url.py
mahmoud: I see someone's a fan of copying.

hyperlink/_url.py
mahmoud: ^

hyperlink/_url.py
mahmoud: Twisted-style single character argument name copied for consistency with URL.

hyperlink/_url.py
mahmoud: Twisted-style camelCase argument name.

Issue #49, opened 2017-11-16
why is str() the same as repr()?


wbolster:
```
>>> url = URL.from_text('https://example.org/api/v2')

>>> str(url)
"URL.from_text('https://example.org/api/v2')"

>>> repr(url)
"URL.from_text('https://example.org/api/v2')"
```

why is `str()` not just the url?

```
https://example.org/api/v2
```

the `requests` library "stringifies" urls (by design). in practice that means that for instance [furl](https://github.com/gruns/furl/) instances can be passed directly. this is not possible with `hyperlink`:

``` python
>>> from hyperlink import URL
>>> import requests

>>> url = URL.from_text('https://example.org/api/v2')
>>> requests.get(url)
[...]
InvalidSchema: No connection adapters were found for 'URL.from_text('https://example.org/api/v2')'
```

mahmoud:
That's a good point, I don't think I'd be opposed to this change. Maybe @Lukasa has an opinion?

mahmoud:
Just merged the fix, this will be released soon! Just need #54 and maybe #56 merged and we'll have a brand new hyperlink 18 :)

wbolster:
awesome, keep up the good work.

PR #55, opened 2017-12-31
add and test URL.__str__, which enables URLs to be passed directly to…

mahmoud:
This `URL.__str__()` enables URLs to be passed directly to libaries which stringify (like requests), as brought up by issue #49.

I've opted to keep it simple and have Python 2's `__str__` (Py3's `__bytes__()`) return the UTF-8 encoded version of `URL.to_text()`. This means that when printed on terminals configured to display UTF-8, the user will get a copy-and-pastable URL.

I also considered making `__str__` return the ASCII-encoded `URL.to_url().to_text()`, but this would have the property of obscuring special characters and possibly falsely-equating URIs and IRIs, contrary to hyperlink's longstanding design.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1343,6 +1343,21 @@ class URL(object):
         """
         return '%s.from_text(%r)' % (self.__class__.__name__, self.to_text())
 
+    def _to_bytes(self):
+        """
+        Allows for direct usage of URL objects with libraries like
+        requests, which automatically stringify URL parameters. See
+        issue #49. Note that some libraries may not handle IRIs well.
+        """
+        return self.to_text().encode('utf8')
+
+    if PY2:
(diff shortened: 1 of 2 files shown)


glyph:
Rather than adding the "some libraries are bad at IRIs" in a private docstring that I guarantee you no one will ever read, if the goal here is `requests` compatibility, perhaps we should apply the `as_url()` transformation unconditionally here?  A higher-fidelity representation is always available via `to_text` if desired.

mahmoud:
The "some libraries are bad at IRIs" is just a hint of a glimmer of something I remember, combined with general suspicion. I'm fine with the always `as_url()` if you are, brb.

markrwilliams:
@glyph https://github.com/twisted/treq/issues/212

Issue #19, opened 2017-06-27
℅ does not encode as a domain name (Python's built-in idna encoding is insufficient)


wsanchez:
I'm not entirely sure this is a bug, but it sure seems like one:

```
>>> from hyperlink import URL
>>> text = u'http://\u2105'
>>> url = URL.fromText(text)
>>> url.asIRI()
URL.from_text('http://℅')
>>> url.asURI()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/wsanchez/Dropbox/Developer/BurningMan/ranger-ims-server/.tox/coverage-py36/lib/python3.6/site-packages/hyperlink/_url.py", line 1070, in to_uri
    fragment=_encode_fragment_part(self.fragment, maximal=True)
  File "/Users/wsanchez/Dropbox/Developer/BurningMan/ranger-ims-server/.tox/coverage-py36/lib/python3.6/site-packages/hyperlink/_url.py", line 861, in replace
    userinfo=_optional(userinfo, self.userinfo),
  File "/Users/wsanchez/Dropbox/Developer/BurningMan/ranger-ims-server/.tox/coverage-py36/lib/python3.6/site-packages/hyperlink/_url.py", line 601, in __init__
    self._host = _textcheck("host", host, '/?#@')
  File "/Users/wsanchez/Dropbox/Developer/BurningMan/ranger-ims-server/.tox/coverage-py36/lib/python3.6/site-packages/hyperlink/_url.py", line 410, in _textcheck
    % (''.join(delims), name, value))
ValueError: one or more reserved delimiters /?#@ present in host: 'c/o'
```

The cause is probably in the IDNA encoding:

```
>>> "℅".encode("idna")
b'c/o'
```

This surprises me, but I might just be ignorant about how this works…?

mahmoud:
Count me among the surprised as well. Looking at the idna package (a possible dependency of hyperlink in the future), it says: `idna.core.InvalidCodepoint: Codepoint U+2105 at position 1 of u'\u2105' not allowed`

It's possible the builtin idna encoding is dropping the ball hard.

wsanchez:
Here's another one:

``
>>> "⁇".encode("idna")
b'??'
```

Note the unencoded `"⁇"` is a single character.

wsanchez:
The built-in IDNA encoder seems fairly weak.

wsanchez:
[...]

PR #56, opened 2018-01-06
Switch to using idna package

mahmoud:
As discussed in issue #19, Python's builtin idna support is outdated and broken. Thankfully the [idna](https://github.com/kjd/idna/) package exists. This PR switches to using that, instead of Python's builtin idna codec.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -579,6 +579,50 @@ def _percent_decode(text, normalize_case=False, subencoding='utf-8',
 
 
 def _decode_host(host):
+    """Decode a host from ASCII-encodable text to IDNA-decoded text. If
+    the host text is not ASCII, it is returned unchanged, as it is
+    presumed that it is already IDNA-decoded.
+
+    Some technical details: _decode_host is built on top of the "idna"
+    package, which has some quirks:
+
+    Capital letters are not valid IDNA2008. The idna package will
+    raise an exception like this on capital letters:
+
+    > idna.core.InvalidCodepoint: Codepoint U+004B at position 1 ... not allowed
+
+    However, if a segment of a host (i.e., something in
+    url.host.split('.')) is already ASCII, idna doesn't perform its
+    usual checks. In fact, for capital letters it automatically
+    lowercases them.
+
+    This check and some other functionality can be bypassed by passing
+    uts46=True to idna.encode/decode. This allows a more permissive and
+    convenient interface. So far it seems like the balanced approach.
+
+    Example output (from idna==2.6):
+
+    >> idna.encode(u'mahmöud.io')
+    'xn--mahmud-zxa.io'
+    >> idna.encode(u'Mahmöud.io')
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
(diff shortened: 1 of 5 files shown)


glyph:
Thinking out loud my anxieties here, rather than a code review:
- I seem to remember that the built-in `idna` encoding manages to encode some things which are officially "invalid" yet recognized by browsers. Am I misremembering?
- Do we care about "zero dependencies"?  Years of being burned by packaging have left their mark; perhaps I shouldn't care, and if you can depend on hyperlink you can depend on its downstreams transitively.
- Why the ceiling pin in `install_requires`?  Is that just for semver compatibility?  Is there a process for noticing that the pin should float a version higher?  I'd be inclined to just leave it out, to avoid conflicting with other software that might need more recent versions.

mahmoud:
@glyph, replies in order:

- I haven't reported/encountered anything like that. More worrisome is that idna encoding certain characters can result in totally-invalid URLs (slashes in the host, like #19). On the positive side, Cory Benfield says they use the idna package for everything. It could be there are old (pre-idna2k8) URLs lying around which might fail to decode, I can make it fallback. Note that I had to kill a doctest but it was clearly an example URL someone typed out, not linked to/used.
- I don't care all that much about no deps. Lots of hyper projects have dependencies. Pretty sure the Twisted `tls` extra pulls in idna anyways.
- The ceiling on install_requires was copied from requests and is basically semver at this point. I'm fine cutting that, too. The Twisted `tls` extra uses `idna!=2.3,>=0.6`.

hyperlink/_url.py
```diff
@@ -1793,3 +1802,38 @@ def parse(url, decoded=True, lazy=False):
...
+* If a segment of a host (i.e., something in url.host.split('.')) is
+already ascii, idna doesn't perform its usual checks. For instance,
+capital letters are not valid idna2008. The package automatically lowercases.
+
+You'll get something like:
+
+> idna.core.InvalidCodepoint: Codepoint U+004B at position 1 ... not allowed
+
+This check and some other functionality can be bypassed by passing
+uts46=True to encode/decode. This allows a more permission and
```
wsanchez: permission ➜ premissive

setup.py
```diff
@@ -29,6 +29,7 @@
       zip_safe=False,
       license=__license__,
       platforms='any',
+      install_requires=['idna>=2.5,<2.7'],
```
wsanchez: I would remove the `<2.7` ceiling

mahmoud:
@wsanchez, all fixed up! The codecov integration on this PR is all froze up, but the report itself says the diff coverage is 100%, so I'm going to merge momentarily. Thanks!

glyph:
Come on codecov do your thing.

glyph:
(Please convince codecov that these lines are all covered somehow, then land.)

mahmoud:
screenshot from 2018-01-07 16-59-46

@glyph yeah it's not gonna update, but the report on the site confirms 100% coverage, see screenshot above.

wsanchez:
@glyph I think I'm not an approved approved because I'm not a committer.

PR #59, opened 2018-01-11
Unicode mixed percent decoding

mahmoud:
Enhance `_percent_decode()` so that it properly decodes percents pairs even when non-ASCII is present, and extended docstring. Remove re-percent-encoding logic from DecodedURL. Now `_percent_decode()` actually does what the name says!

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -523,18 +523,22 @@ def _percent_decode(text, normalize_case=False, subencoding='utf-8',
     u'abc def'
 
     Args:
-       text (unicode): The ASCII text with percent-encoding present.
+       text (unicode): Text with percent-encoding present.
        normalize_case (bool): Whether undecoded percent segments, such
           as encoded delimiters, should be uppercased, per RFC 3986
           Section 2.1. See :func:`_decode_path_part` for an example.
+       subencoding (unicode): The name of the encoding underlying the
+          percent-encoding. Pass `False` to get back bytes.
+       raise_subencoding_exc (bool): Whether an error in decoding the bytes
+          underlying the percent-decoding should be raised.
 
     Returns:
-       unicode: The percent-decoded version of *text*, with UTF-8
-         decoding applied.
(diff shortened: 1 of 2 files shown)


hyperlink/_url.py
```diff
@@ -523,18 +523,22 @@ def _percent_decode(text, normalize_case=False, subencoding='utf-8',
...
     Returns:
-       unicode: The percent-decoded version of *text*, with UTF-8
-         decoding applied.
+       unicode: The percent-decoded version of *text*, with decoding
+         applied, unless `subencoding=False` which returns bytes.
 
     """
     try:
-        quoted_bytes = text.encode("ascii")
+        quoted_bytes = text.encode(subencoding or 'utf-8')
```
markrwilliams: I think this is the actual API surface, with `ascii` as a stand in for an arbitrary non-UTF-8 encoding:

| `text`                       | `subencoding`       | `quoted_bytes`           | return value                 |
|------------------------------|---------------------|--------------------------|------------------------------|
| `<unicode>`                  | `"utf-8"` (default) | `<bytes, UTF-8 encoded>` | `<unicode, percent-encoded>` |
| `<unicode>`                  | `"ascii"`           | `UnicodeEncodeError`     | `<unicode, unaltered>`       |
| `<unicode, ASCII congruent>` | `"ascii"`           | `<bytes, ASCII>`         | `<bytes, ASCII>`             |
| `<unicode, charmap?>`        | `False`             | `<bytes, UTF-8 encoded>` | `<bytes, UTF-8 encoded>`     |

The last case seems weird - I want `bytes` back because I know there isn't actually a "subencoding" (and I ended up with `unicode` presumably because somewhere else there's something like `.decode("charmap")`), but instead I get UTF-8 encoded bytes back.  What then is the purpose of `subencoding=False` if it always implies UTF-8?

I think splitting this line up into an explicit `if` statement will make the situation clearer, and also make it easier to write tests that cover each case.

Issue #61, opened 2018-02-17
normalizing "free radical" percent signs


glyph:
I think that
```
>>> hyperlink.URL(path=['%%%']).normalize()
URL.from_text('%%%')
```
ought to be giving me `%25%25%25` ?

I haven't managed to dredge up a spec reference for this, but a % in the path without 2 hex digits after it seems like it ought to just be quoted.

mahmoud:
Yeah, I agree this would be nice behavior. Spec says free percents should always be encoded, but in a quick test, neither Firefox nor Chrome do this:

```
$ nc -l -p 9999

# firefox

GET /%%%/%%% HTTP/1.1
Host: localhost:9999
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:58.0) Gecko/20100101 Firefox/58.0
...

# chromium
GET /%%%/%%% HTTP/1.1
Host: 127.0.0.1:9999
Connection: keep-alive
Cache-Control: max-age=0
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/64.0.3282.167 Chrome/64.0.3282.167 Safari/537.36
...
```

So, to that expectation, `.normalize()` does seem like the right place to put this sort of thing. I'll look into it now.

mahmoud:
@glyph, got a PR for ya (#62), as mentioned above. Thanks for the initial report and thanks in advance if you're able to review.

PR #62, opened 2018-02-24
Add encode_stray_percents functionality.

mahmoud:
Per #61. This PR adds a new argument to `_percent_decode()`, its variants, and `URL.normalize()`. 

This option (disabled by default, and enabled for normalize), exists to turn unmatched `%` characters to their encoded equivalent, correcting an otherwise improperly encoded URL. Major browsers don't do this by default, so it's currently limited to the explicit normalize step.


mahmoud:
@glyph, you got it, `normalize()` now takes a "percents" argument.

Issue #66, opened 2018-07-07
DecodedURL.click() only permits unicode or EncodedURL argument


agoose77:
Calling `DecodedURL.click(url)` raises a `TypeError` for `DecodedURL` arguments. A solution to this might be to check for `DecodedURL` and pass `url._url` into the wrapped click method, or to define a common base class for the two URL types and type-check that. I'm not sure whether interchanging between EncodedURL and DecodedURL is valid semantically, however.

mahmoud:
Good catch! Definitely missed adding that functionality in the first round. I added a test/fix on a branch for this issue in particular, and I'm going to do a quick once-over to make sure other methods aren't similarly affected. Thanks!

PR #67, opened 2018-07-07
Augment DecodedURL.click() to support DecodedURL arguments

mahmoud:
Pretty straightforward test + fix. Addresses #66.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1594,8 +1594,8 @@ class DecodedURL(object):
 
     Args:
        url (URL): A :class:`URL` object to wrap.
-       lazy (bool): Whether to pre-decode all parts of the URL to
-           check for validity. Defaults to True.
+       lazy (bool): Set to True to avoid pre-decode all parts of the
+           URL to check for validity. Defaults to False.
 
     """
     def __init__(self, url, lazy=False):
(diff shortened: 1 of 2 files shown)


hyperlink/_url.py
```diff
@@ -1594,8 +1594,8 @@ class DecodedURL(object):
 
     Args:
        url (URL): A :class:`URL` object to wrap.
-       lazy (bool): Whether to pre-decode all parts of the URL to
-           check for validity. Defaults to True.
+       lazy (bool): Set to True to avoid pre-decode all parts of the
```
Julian: Minor grammar error here, should now be `avoid pre-decoding`, but the sentence is still a bit awkward compared to the original. Maybe: "If `True`, decode all parts of the URL immediately to check that they are valid. Otherwise when `False`, do so lazily as needed, so an error may be raised <whenever that is true>."

hyperlink/_url.py
```diff
@@ -1640,6 +1640,8 @@ def to_iri(self, *a, **kw):
 
     def click(self, href=u''):
         "Return a new DecodedURL wrapping the result of :meth:`~hyperlink.URL.click()`"
+        if isinstance(href, DecodedURL):
```
Julian: I guess the answer is "no", but there's no URL interface here to rely on?

Julian: Oh never mind, I see, this is in `DecodedURL` itself.

glyph:
Hey @mahmoud do you want to merge this?

Issue #70, opened 2019-03-03
Extend `URL().remove()` to support removing a single key-value pair


alexwlchan:
I have some URLs that use multiple query parameters to track filters, e.g.:

```
/documents?filter=apple&filter=banana&filter=coconut
```

If I parse this with `URL()`, I can use `remove("filter")` to remove all the instances of this query parameter, but removing just the "banana" filter is a little more fiddly.

How would you feel about extending `remove()` to support passing an optional query parameter value to remove, so you only remove that one pair (if it exists)?

Something like:

```python
def remove(self, name, value=None):
    if value is None:
        new_query = ((k, v) for (k, v) in self.query if k != name)
    else:
        new_query = ((k, v) for (k, v) in self.query if k != name and v != name)
    return self.replace(query=new_query)
```

mahmoud:
Sounds good to me! I think I see a bug at the end of the else block there, but I get the picture enough for a PR :)

mahmoud:
@alexwlchan care to review #71?

PR #71, opened 2019-03-26
Add value kwarg to URL.remove()

mahmoud:
Allows filtering for specific values to remove, per #70.

### DIFF
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1802,10 +1802,22 @@ class DecodedURL(object):
         q[idx:idx] = [(name, value)]
         return self.replace(query=q)
 
-    def remove(self, name):
+    def remove(self, name, value=_UNSET, limit=None):
         "Return a new DecodedURL with query parameter *name* removed."
-        return self.replace(query=((k, v) for (k, v) in self.query
-                                   if k != name))
+        if limit is None:
+            if value is _UNSET:
+                nq = [(k, v) for (k, v) in self.query if k != name]
+            else:
+                nq = [(k, v) for (k, v) in self.query if not (k == name and v == value)]
+        else:
+            nq, removed_count = [], 0
+            for k, v in self.query:
+                if k != name and (value is _UNSET or v != value) or removed_count >= limit:
+                    nq.append((k, v))
+                else:
+                    removed_count += 1
+
+        return self.replace(query=nq)
 
     def __repr__(self):
         cn = self.__class__.__name__
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1608,11 +1608,12 @@ class URL(object):
                 nq = [(k, v) for (k, v) in self.query if not (k == name and v == value)]
         else:
             nq, removed_count = [], 0
+
             for k, v in self.query:
-                if k != name and (value is _UNSET or v != value) or removed_count >= limit:
-                    nq.append((k, v))
+                if k == name and (value is _UNSET or v == value) and removed_count < limit:
+                    removed_count += 1  # drop it
                 else:
-                    removed_count += 1
+                    nq.append((k, v))  # keep it
 
         return self.replace(query=nq)
 
@@ -1842,10 +1843,10 @@ class DecodedURL(object):
         else:
             nq, removed_count = [], 0
             for k, v in self.query:
-                if k != name and (value is _UNSET or v != value) or removed_count >= limit:
-                    nq.append((k, v))
+                if k == name and (value is _UNSET or v == value) and removed_count < limit:
+                    removed_count += 1  # drop it
                 else:
-                    removed_count += 1
+                    nq.append((k, v))  # keep it
 
         return self.replace(query=nq)
 
--- a/hyperlink/_url.py
+++ b/hyperlink/_url.py
@@ -1560,7 +1560,7 @@ class URL(object):
         """
         return [value for (key, value) in self.query if name == key]
 
-    def remove(self, name):
+    def remove(self, name, value=_UNSET):
         """Make a new :class:`URL` instance with all occurrences of the query
         parameter *name* removed. No exception is raised if the
         parameter is not already set.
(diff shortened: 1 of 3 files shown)


hyperlink/_url.py
```diff
@@ -1560,7 +1560,7 @@ def get(self, name):
         """
         return [value for (key, value) in self.query if name == key]
 
-    def remove(self, name):
+    def remove(self, name, value=_UNSET):
```
alexwlchan: You need to update the Args list and the description for the new parameter, I think?

hyperlink/_url.py
```diff
@@ -1560,20 +1560,26 @@ def get(self, name):
...
+               both name and value.
 
         Returns:
             URL: A new :class:`URL` instance with the parameter removed.
-
         """
-        return self.replace(query=((k, v) for (k, v) in self.query
-                                   if k != name))
+        if value is _UNSET:
+            nq = [(k, v) for (k, v) in self.query if k != name]
```
glyph: This was previously, extremely intentionally, a generator expression so that the `replace` call wouldn't need to build a temporary list and then replace it with a tuple, it could avoid the intermediate allocation and go straight to the tuple.  Can you make these genexps instead?

alexwlchan:
> I'd really like to see an explicit decision (i.e.: documentation, tests) about what to do in the case of e.g. `?foo=1&foo=1&foo=1`. My feeling would be that it should only remove _one_ pair (e.g. `?foo=1&foo=1`) but really I'm okay with either option as long as it's explicit.

I have no strong opinion, but will note that if we pick the “only remove one pair” option, we should be explicit about which one gets removed. e.g.

`?foo=1&bar=2&foo=1&foo=1`
 ~> `?bar=2&foo=1&foo=1` or
 ~> `?foo=1&bar=2&foo=1`

mahmoud:
> Before this lands, I'd really like to see an explicit decision (i.e.: documentation, tests) about what to do in the case of e.g. `?foo=1&foo=1&foo=1`. My feeling would be that it should only remove _one_ pair (e.g. `?foo=1&foo=1`) but really I'm okay with either option as long as it's explicit.

I thought it was pretty explicit from the docs and tests that it is designed to remove all instances of that name and value (when both are specified).

The right way to do a single removal would be to add and use a `limit` kwarg, like other APIs in the python-verse. I can add that if you say you have a use case for it. Otherwise I prefer the incremental enhancement of just the value filter.

hyperlink/_url.py
mahmoud: Did you bench that?

Lists (and list comps) are faster than genexps for low-level operations like this, even if there's an extra intermediate list being constructed:

```
$ python -m timeit -s "x = [1, 2, 3, 4, 5, 6, 7]" "tuple([y for y in x])"
1000000 loops, best of 3: 0.319 usec per loop
$ python -m timeit -s "x = [1, 2, 3, 4, 5, 6, 7]" "tuple(y for y in x)"
1000000 loops, best of 3: 0.631 usec per loop
$ python -m timeit -s "x = list(range(1000))" "tuple([y for y in x])"
100000 loops, best of 3: 19 usec per loop
$ python -m timeit -s "x = list(range(1000))" "tuple(y for y in x)"
10000 loops, best of 3: 36.7 usec per loop
```

(yes on py2 and py3) 

URLs are a pretty clear-cut case of being weighted toward short lists. Most clients have URL length limits and also the limits of clipboards and human readability.

I've tried in the past not to generate extra lines of diff in such a closely-scrutinized library, but I'm not going to unlearn my micro-optimization muscle memory. I'll probably change others when I touch those lines, and request others do the same.

If you have other such cases of extreme intention outside of hyperlink, I would revisit them to see if you're losing cycles. Generators are like a once a month thing, outside of inlineCallbacks and other nice high-level stuff. :v:

mahmoud:
@glyph / @alexwlchan comments responded, docstrings updated, added a `limit` parameter, DecodedURL .remove() API updated to match, feel free to review!

hyperlink/_url.py
```diff
@@ -1585,20 +1585,36 @@ def get(self, name):
...
-                                   if k != name))
+        if limit is None:
+            if value is _UNSET:
+                nq = [(k, v) for (k, v) in self.query if k != name]
+            else:
+                nq = [(k, v) for (k, v) in self.query if not (k == name and v == value)]
+        else:
+            nq, removed_count = [], 0
+            for k, v in self.query:
+                if k != name and (value is _UNSET or v != value) or removed_count >= limit:
```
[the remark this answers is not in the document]
mahmoud: Yeah, agreed that the logic there is a bit more than fits into the ole discrete buffers. But first off, the spec says that the `name` and `value` should _both_ match, so the first `or` being an `and` is intentional. We're not supporting the use case of removing all query params with a given value.

Still, I inverted the structure for a net reduction in boolean operators (got rid of some nots) and I think it's more readable now:

```python
            nq, removed_count = [], 0
            for k, v in self.query:
                if k == name and (value is _UNSET or v == value) and removed_count < limit:
                    removed_count += 1  # drop it
                else:
                    nq.append((k, v))  # keep it

        return self.replace(query=nq)
```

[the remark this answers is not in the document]
alexwlchan: Oops. 😆

The new version is much clearer, thanks!

hyperlink/_url.py
mahmoud: @glyph my turn to bump. ;) I've resolved Alex's comments, and addressed the above, so I think this is ready to merge. Y/n?

hyperlink/_url.py
```diff
@@ -1585,20 +1585,37 @@ def get(self, name):
...
         """Make a new :class:`URL` instance with all occurrences of the query
-        parameter *name* removed. No exception is raised if the
+        parameter *name* removed, or, if *value* is set, parameters
+        matching *name* and *value*. No exception is raised if the
         parameter is not already set.
 
         Args:
             name (unicode): The name of the query parameter to remove.
+            value (unicode): Optional value to additionally filter
+               on. Setting this removes query parameters which match
```
glyph: If you could drop the word "all" in here, that would I think fully address my concern.

mahmoud: done!

mahmoud:
This feature, along with hyperlink 19.0.0 now up on PyPI, thanks to you both! Enjoy!

PR #72, opened 2019-09-18
Add PyPI license classifier

RazerM:
This makes it easier to introspect licenses for installed packages (e.g. with [pip-licenses](https://pypi.org/project/pip-licenses/))


mahmoud:
Good catch, thanks!

PR #73, opened 2019-10-18
Add Python 3.7 to build matrix, re-work build config

wsanchez:
Add Python 3.7.
The (pretty old) version of `pytest` didn't like 3.7, so I updated the dependencies.
And it turns out that newest `pytest` dropped Python 2 and 3.5.

Enter rabbit hole, so I started with a working config similar to that used by Klein and some other projects, which meant:

`coverage`:
 * Don't use `coverage run` and let `pytest-cov` do the heavy lifting.
 * Rename .tox-coveragerc ➜ .coveragerc and greatly simplify it, because `pytest-cov` doesn't seem to need as more hooey.

`tox`:
 * Use `pytest` 4.6 for older Pythons and 5.2 for the rest.
 * Itemize explicit `basepython` values
 * Pass more environment variables to Codecov per Codecov docs
 * Set `skip_missing_interpreters=True` if on a tty, so less noise if you don't install every python, but still fails if they are missing in CI.
 * Also add pypy3 to the matrix

`appveyor`:
 * Tweaks to catch up with too config changes.


wsanchez:
This requires updates to pytest at least… will update dependencies.

wsanchez:
Codecov not coming back doesn't appear to be the service's fault (for once).

mahmoud:
This looks great! And really close to working. I wonder what's happening on appveyor that it's giving a green check but the windows compat code isn't getting hit/reported on codecov.

wsanchez:
AppVeyor seems to want a Codecov token to submit with (Travis gets this for free), but I'm not sure where it was getting that token from before.

wsanchez:
Yeah, look's like that's the only blocker here; everything passed except for coverage for some windows-specific code.

mahmoud:
Hmm, pretty sure Appveyor hasn't needed a token for some years now.

> If you have a public project on TravisCI, CircleCI, AppVeyor, or Azure Pipelines an upload token is not required.

(from https://docs.codecov.io/docs/about-the-codecov-bash-uploader)

Maybe there's something in the diff, I'm taking a look.

wsanchez:
> Hmm, pretty sure Appveyor hasn't needed a token for some years now.

Yup, just read the same thing a moment ago. :-)

I suspect that Appveyor is sticking the token into an environment variable that tax is not passing through.  (The prior config was running codecov directly, not via tox.)

wsanchez:
[Found it](https://github.com/codecov/codecov-python/blob/master/codecov/__init__.py#L419)

wsanchez:
OK, added APPVEYOR_* and links to some documentation in comments.
Also added GITHUB_* and docs so we don't have to dig for that if we want to try GitHub Actions later.

wsanchez:
Yay! OK I've got a bit of cleanup, then will be ready for review.

wsanchez:
OK.  All set.

tox.ini
```diff
@@ -1,26 +1,121 @@
...
+        APPVEYOR_REPO_BRANCH APPVEYOR_REPO_COMMIT \
+        APPVEYOR_REPO_COMMIT_AUTHOR_EMAIL APPVEYOR_REPO_COMMIT_MESSAGE_EXTENDED
+
+
+##
+# Packaging
+##
 
 [testenv:packaging]
-changedir = {toxinidir}
```
mahmoud: slightly concerned about dropping `changedir` here, because this prevents python from loading things in from the current path, as described here: https://blog.ganssle.io/articles/2019/08/test-as-installed.html

tox.ini
```diff
@@ -1,26 +1,121 @@
 [tox]
-envlist = py26,py27,py34,py35,py36,pypy,coverage-report,packaging
+
+envlist =
+    test-py{26,27,34,35,36,37,py,py3}
+    coverage_report
+    packaging
+
+skip_missing_interpreters = {tty:True:False}
```
mahmoud: Didn't know this was a feature, nice. Syntax for the value tripped me out for a sec, but not bad. :)

.tox-coveragerc
```diff
@@ -1,14 +0,0 @@
-[run]
-branch = True
-source =
-    hyperlink
-    ../hyperlink
-omit =
-    */flycheck_*
```
mahmoud: I may want to get this omit directive back just for me and my fellow emacs flycheck users. :)

tox.ini
wsanchez: Yeah, in Klein I used this plus a wrapper script:
```python
skip_missing_interpreters = {env:TOX_SKIP_MISSING_INTERPRETERS:True}
```
But then I saw they added this `tty` thing, and that seems even better.

tox.ini
wsanchez: For the packaging test, we aren't running the python code; we're running `setup.py`.  And `toxinidir` is already the default working directory.

For the tests… I was planning on following this up with the addition of a `src/` directory, which I think is sufficient for the stated purpose (and works with or without `tox`) and a known best practice.  I can put the `changedir` back, but I don't know what that would really add, and despite making the tax config a lot bigger already, I do prefer not to add things that aren't needed.

That said, I'm not all that invested in it if you think it's better to put it back.

.tox-coveragerc
wsanchez: Ah.  I had no idea what that was for, and got it in my head that it must be some leftover thing and then forgot to ask.  I'll stick it back in.

wsanchez: done

tox.ini
wsanchez: On further reflection, if I remove this, it should happen *after* the `src/` thing, so I'll stick it back in.

wsanchez: Right… that also requires other path hooey elsewhere.  How about I just do the src/ thing?

appveyor.yml
```diff
@@ -2,49 +2,40 @@
...
+    - PYTHON: "C:\\Python37-x64"
+      TOX_ENV: "test-py37,codecov"
 
 
 init:
+  - set OS=WINDOWS
   - set PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
   - "git config --system http.sslcainfo \"C:\\Program Files\\Git\\mingw64\\ssl\\certs\\ca-bundle.crt\""
   - "%PYTHON%/python -V"
   - "%PYTHON%/python -c \"import struct;print(8 * struct.calcsize(\'P\'))\""
```
wsanchez: I wonder what the red highlighting in the diff here is about…

mahmoud: I think it thinks something's off with the escaping. I've seen it around from time to time, doesn't seem to be correlated with actual breakage so I haven't given it much more thought lol

tox.ini
mahmoud: Haha src thing is fine by me, I think it's gotten popular enough now that most people can figure it out. And when in doubt there's links to send folks.

mahmoud:
Changes LGTM to me, so I'm gonna merge. This has been a great effort and a great outcome, thank you @wsanchez!

PR #74, opened 2019-10-19
Use assertEqual(), not assertEquals(), which is deprecated.

wsanchez:


mahmoud:
Also, I appreciate all the work you're doing, looking forward to #73 thanks a bunch in advance!

PR #75, opened 2019-10-21
Add Flake8 to Tox config.

wsanchez:


src/hyperlink/_url.py
```diff
@@ -29,16 +29,15 @@
 except ImportError:  # Python 2
     from collections import Mapping
 
-# Note: IDNAError is a subclass of UnicodeError
-from idna import encode as idna_encode, decode as idna_decode, IDNAError
```
wsanchez: IDNAError was unused

src/hyperlink/_url.py
```diff
@@ -29,16 +29,15 @@
...
+from idna import encode as idna_encode, decode as idna_decode
 
 
 if inet_pton is None:
     # based on https://gist.github.com/nnemkin/4966028
     # this code only applies on Windows Python 2.7
     import ctypes
 
-    class _sockaddr(ctypes.Structure):
+    class _sockaddr(ctypes.Structure):  # noqa: N801
```
wsanchez: `N801` is "you should use ClassNameCaps, yo", which… actually I guess we can do that here?

wsanchez: The leading `"_"` is probably unnecessary here, since the whole module is private.

src/hyperlink/_url.py
```diff
@@ -742,7 +757,7 @@ def parse_host(host):
 
 
 class URL(object):
-    """From blogs to billboards, URLs are so common, that it's easy to
+    r"""From blogs to billboards, URLs are so common, that it's easy to
```
wsanchez: Made this a raw string because there are `\` characters in there which are not actually trying to escape the next character and chasing those to `\\` actually wrecks the string's mojo.

tox.ini
```diff
@@ -59,6 +60,68 @@ commands =
...
+basepython = python3.7
+
+skip_install = True
+
+# Pin pydocstyle to version 3: see https://gitlab.com/pycqa/flake8-docstrings/issues/36
+deps =
+    flake8-bugbear==19.8.0
+    #flake8-docstrings==1.4.0
+    #flake8-import-order==0.18.1
+    #flake8-pep3101==1.2.1
```
wsanchez: This PR is already pretty noisy, so not piling on with these.

src/hyperlink/test/test_url.py
```diff
@@ -10,9 +10,7 @@
 
 from .common import HyperlinkTestCase
 from .. import URL, URLParseError
-# automatically import the py27 windows implementation when appropriate
-from .. import _url
```
wsanchez: `_url` is not used, and tests are passing, but the comment makes me nervous here.
@mahmoud: perhaps you could double-check this bit…

src/hyperlink/_url.py
wsanchez: Changed to `SockAddr`

src/hyperlink/test/test_url.py
mahmoud: The comment and the import are from almost a year apart according to blames. The comment relates to inet_pton testing, and really doesn't add much. If tests pass we're fine :)

src/hyperlink/_url.py
mahmoud: as long as it looks fine in the docs :) We'll see what sphinx/readthedocs does with it.

src/hyperlink/_url.py
mahmoud: I think I just left it cased that way as an homage more than anything. It's fine titlecased.

src/hyperlink/_url.py
mahmoud: Yeah I think at one point there may have been thoughts of catching and reraising, but now we just let them through.

tox.ini
mahmoud: fair :)

wsanchez:
> Overall looks good to me! The long line reformatting gives me thoughts of possibly blackifying the codebase. any experience with that? It'll possibly wreck the blame with the history back to twisted land...

No, but I've been itching to dive into it, so I might give it a go.

PR #76, opened 2019-10-21
Turn on flake8-import-order.

wsanchez:


mahmoud:
Hmm, not sure how I feel about this one. I may be one of the odd ones in that, I care about import formatting (grouping stdlib, extlib, and internal imports is great), but over the years I've not found alphabetizing to help, and generally I think the code on the right looks worse?

Again, not a big deal, but curious if you've found it to help much, if at all?

wsanchez:
What I find is that it eliminates variation between developers, which is useful when you have >1 developer actively working on the code.  On the flip side, I don't know how the code on the right looks worse…  I find that the re-sorting of the import lines, including the alphabetizing, makes it read better, and sorting both the lines and symbols within the import lines makes it easier to find imports.

All that said, if you dislike it, we can drop this PR.  It's not functionally different, and the above is not earth-shattering.

PR #77, opened 2019-10-21
Enable flake8-pep3101.

wsanchez:


wsanchez:
The Codecov-patch fail above is because some previously untested code remains untested, but I touched it to change `%`-formatting to `str.format`.

While I'm all in favor of adding tests, I think in this case, that's an orthogonal issue best addressed separately.  Alternatively, we could revert the uncovered changes and disable that warning for now, but that doesn't seem better to me.

@mahmoud:  Not sure how you feel about that.

mahmoud:
At its heart, flake8-pep3101 is flawed in thinking there's "old" string formatting. To the best of my knowledge, %-formatting is still supported, is more succinct, and is about twice as fast as `.format()`. I can see switching over to f-strings if/when we go py3-only, but unless I missed a memo, I'm pretty ambivalent on this change.

wsanchez:
OK, the nail in the coffin for me with `%`-formatted strings is the tuple-or-not argument they take, which should always take the form of a tuple, lest you pass in a symbol that is a tuple and suddenly some other thing is happening, so I avoid them.

That said, I try to avoid writing new code for <3.6, so I'm usually using f-strings.

PR #78, opened 2019-10-21
Fix a typo in the docs

hristog:


mahmoud:
TODO's thankfully not really docs, but thank you all the same!

PR #82, opened 2019-11-01
Add more type hints

wsanchez:
This completes the type hints for the project.


src/hyperlink/_url.py
```diff
@@ -749,37 +775,46 @@ class URL(object):
...
+            key-value pairs.
+        fragment: The fragment part of the URL.
+        rooted: Whether or not the path begins with a slash.
+        userinfo: The username or colon-separated username:password pair.
+        uses_netloc: Indicates whether two slashes appear between the scheme
+            and the host (``http://eg.com`` vs. ``mailto:e@g.com``).
+            Set automatically based on scheme.
+
+    All of these parts are also exposed as read-only attributes of URL
+    instances, along with several useful methods.
```
wsanchez: I removed the type information from the docstring, since (a) some of it is actually inaccurate and (b) it's redundant given the type hints.

src/hyperlink/_url.py
```diff
@@ -749,37 +775,46 @@ class URL(object):
...
+        port=None,         # type: Optional[int]
+        rooted=None,       # type: Optional[bool]
+        userinfo=u"",      # type: Text
+        uses_netloc=None,  # type: Optional[bool]
+    ):
+        # type: (...) -> None
         if host is not None and scheme is None:
             scheme = u'http'  # TODO: why
-        if port is None:
+        if port is None and scheme is not None:
```
wsanchez: The added check is because the following call to `.get` expects `Text`, not `Optional[Text]`, and `None` isn't in the dict, so the result would be `None`, which is already what `port` is set to here, so there you go.

src/hyperlink/_url.py
mahmoud: Maybe we should upgrade/setup hyperlink's sphinx to include those in the docs somehow? (This is [Napoleon doing Google-style docs](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html), btw.)

wsanchez: I don't know enough about Sphinx to do that, but I agree that it would be cool if it could be set up that way.

mahmoud: Sure, I'll take a swing at that, then. Gotta do something around here, after all! :)

wsanchez: :-) no worries; we pitch in when we have cycles and I just happen to have a bust of energy for this right now.  I haven't even gotten to the think I want to work on yet!

But if you can figure out how to get Sphinx to show the type hints, that would really be awesome.  If not, I thing I can go through and put these back with more accurate (but, sadly, duplicate) info.

src/hyperlink/_url.py
```diff
@@ -19,36 +19,50 @@
...
+QueryParameters = Union[
+    Mapping[Text, Optional[Text]], QueryPairs,
+    Sequence[Tuple[Text, Optional[Text]]],
+]
+T = TypeVar('T')
 
 
 # from boltons.typeutils
-def make_sentinel(name='_MISSING', var_name=None):
+def make_sentinel(name='_MISSING', var_name=""):
```
wsanchez: Made the default `""` here so we don't have to make `var_name` an `Optional`.

src/hyperlink/_url.py
```diff
@@ -219,8 +243,14 @@ def _encode_schemeless_path_part(text, maximal=True):
...
-def _encode_path_parts(text_parts, rooted=False, has_scheme=True,
-                       has_authority=True, joined=True, maximal=True):
+def _encode_path_parts(
+    text_parts,          # type: Sequence[Text]
+    rooted=False,        # type: bool
+    has_scheme=True,     # type: bool
+    has_authority=True,  # type: bool
+    maximal=True,        # type: bool
+):
+    # type: (...) -> Sequence[Text]
```
wsanchez: Note that I removed the `joined` argument here.
Keeping it would mean that we're return either a sequence of strings or a string, which would require casting elsewhere.  It's easy enough for the caller to join the parts if needed, which IIRC I had to only fix in one place.

src/hyperlink/_url.py
```diff
@@ -528,24 +577,19 @@ def _percent_decode(text, normalize_case=False, subencoding='utf-8',
...
-       unicode: The percent-decoded version of *text*, decoded by
-         *subencoding*, unless `subencoding=False` which returns bytes.
-
+        The percent-decoded version of *text*, decoded by *subencoding*.
     """
     try:
-        quoted_bytes = text.encode(
-            'utf-8' if subencoding is False else subencoding
-        )
+        quoted_bytes = text.encode(subencoding)
```
wsanchez: Removed code path for `subencoding=`False`.
It was not used by any callers, and it required that we return a `Union[str, bytes]` where returning `Text` all the time avoids casting elsewhere.

src/hyperlink/_url.py
```diff
@@ -579,8 +623,6 @@ def _percent_decode(text, normalize_case=False, subencoding='utf-8',
 
     unquoted_bytes = b''.join(res)
 
-    if subencoding is False:
-        return unquoted_bytes
```
wsanchez: Removed code path for subencoding=False.  See above.

src/hyperlink/_url.py
```diff
@@ -701,22 +747,24 @@ def parse_host(host):
...
+            family = AF_INET6  # type: Optional[AddressFamily]
     else:
-        family = socket.AF_INET
+        try:
+            inet_pton(AF_INET, host)
+        except (socket.error, UnicodeEncodeError):
+            family = None  # not an IP
+        else:
+            family = AF_INET
+
```
wsanchez: Some re-arranging here so that the only `return` statement is below, and this code just sets those values.

src/hyperlink/_url.py
```diff
@@ -827,14 +884,16 @@ def __init__(self, scheme=None, host=None, path=(), query=(), fragment=u'',
         return
 
     def get_decoded_url(self, lazy=False):
+        # type: (bool) -> DecodedURL
         try:
             return self._decoded_url
         except AttributeError:
-            self._decoded_url = DecodedURL(self, lazy=lazy)
+            self._decoded_url = DecodedURL(self, lazy=lazy)  # type: DecodedURL
```
wsanchez: I think it is really weird that this type hint is needed; seems like a `mypy` bug.

src/hyperlink/_url.py
```diff
@@ -19,36 +19,50 @@
...
+    from collections.abc import Mapping as MappingABC
 except ImportError:  # Python 2
-    from collections import Mapping
+    from collections import Mapping as MappingABC
 
 from idna import encode as idna_encode, decode as idna_decode
 
 
 PY2 = (sys.version_info[0] == 2)
-unicode = type(u'')
```
wsanchez: `typing.Text` is `unicode` on Py2 and `str` on Py3.  Since we're importing that now, we don't need this.

src/hyperlink/_url.py
```diff
@@ -1095,22 +1170,19 @@ def from_text(cls, text):
...
             and IPv6 values in the host portion of the URL.
-
         """
         um = _URL_RE.match(_textcheck('text', text))
-        try:
-            gs = um.groupdict()
-        except AttributeError:
+        if um is None:
             raise URLParseError('could not parse url: %r' % text)
+        gs = um.groupdict()
```
wsanchez: If `um` isn't `None`, it shouldn't raise `AttributeError` here, so this now just checks if `um` is `None`.  That avoids some casting.

src/hyperlink/_url.py
```diff
@@ -1095,22 +1170,19 @@ def from_text(cls, text):
...
         au_text = gs['authority'] or u''
         au_m = _AUTHORITY_RE.match(au_text)
-        try:
-            au_gs = au_m.groupdict()
-        except AttributeError:
+        if au_m is None:
             raise URLParseError(
                 'invalid authority %r in url: %r' % (au_text, text)
             )
+        au_gs = au_m.groupdict()
```
wsanchez: If `au_m` isn't `None`, it shouldn't raise `AttributeError` here, so this now just checks if `au_m` is `None`. That avoids some casting.

src/hyperlink/_url.py
```diff
@@ -1122,7 +1194,7 @@ def from_text(cls, text):
         port = au_gs['port']
         if port is not None:
             try:
-                port = int(port)
+                port = int(port)  # type: ignore[assignment] FIXME, also below
```
wsanchez: When we set port a few lines up, it's a string, and now we're setting it to an `int`.
That should be cleaned up, as currently port is sometimes a string when it is passed to `cls()`, and I think that's incorrect.
I made some attempts are fixing it, and kept running into problems, so it probably should go in a separate PR.

src/hyperlink/_url.py
```diff
@@ -1410,24 +1504,26 @@ def to_text(self, with_password=False):
...
-                                  has_scheme=bool(scheme),
-                                  has_authority=bool(authority),
-                                  maximal=False)
+        path = "/".join(_encode_path_parts(
+            self.path,
+            rooted=self.rooted,
+            has_scheme=bool(scheme),
+            has_authority=bool(authority),
+            maximal=False
+        ))
```
wsanchez: `_encode_path_parts` no longer returns a string by default (it returns a sequence of parts).

src/hyperlink/_url.py
```diff
@@ -1705,59 +1822,85 @@ def child(self, *segments):
...
 
     @property
     def path(self):
+        # type: () -> Sequence[Text]
         try:
-            return self._path
+            return cast(
+                Tuple[Text, ...],
+                self._path  # type: ignore[has-type] can't determine
+            )
```
wsanchez: Not sure why `mypy` can't figure out the type of `self._path` here, especially given that I'm telling it with the `cast` here.
There are a few more of these below.

src/hyperlink/_url.py
```diff
@@ -1793,46 +1959,61 @@ def replace(self, scheme=_UNSET, host=_UNSET, path=_UNSET, query=_UNSET,
...
+                    for x in (k, v)
+                )
+                for k, v in iter_pairs(query)
+            ))
         if userinfo is not _UNSET:
             if len(userinfo) > 2:
                 raise ValueError('userinfo expected sequence of ["user"] or'
-                                 ' ["user", "password"], got %r' % userinfo)
-            userinfo = u':'.join([_encode_reserved(p) for p in userinfo])
+                                 ' ["user", "password"], got %r' % (userinfo,))
```
wsanchez: Here's an example of why `%`-formatting sucks.  If `userinfo` is a tuple, the prior code poops.

src/hyperlink/test/test_decoded_url.py
```diff
@@ -178,10 +185,6 @@ def test_twisted_compat(self):
         assert 'asText' not in dir(durl)
         assert durl.to_text() == durl.asText()
 
-    def test_percent_decode_bytes(self):
-        # type: () -> None
-        assert _percent_decode('%00', subencoding=False) == b'\0'
-
```
wsanchez: `subencoding=False` handling was removed.

src/hyperlink/test/test_decoded_url.py
```diff
@@ -196,9 +199,6 @@ def test_percent_decode_mixed(self):
         with self.assertRaises(UnicodeDecodeError):
             _percent_decode('abcdé%C3éfg', raise_subencoding_exc=True)
 
-        # check that getting raw bytes works ok
-        assert _percent_decode('a%00b', subencoding=False) == b'a\x00b'
-
```
wsanchez: `subencoding=False` handling was removed.

wsanchez:
OK this is ready for review.  I'd put in comments in the diffs where the change may not be obvious.

src/hyperlink/test/test_url.py
```diff
@@ -883,41 +883,41 @@ def check(param, expectation=defaultExpectation):
...
-            url.child(Unexpected())
+            url.child(cast(Text, Unexpected()))
         assertRaised(raised, defaultExpectation, "path segment")
         with self.assertRaises(TypeError) as raised:
-            url.sibling(Unexpected())
+            url.sibling(cast(Text, Unexpected()))
         assertRaised(raised, defaultExpectation, "path segment")
         with self.assertRaises(TypeError) as raised:
-            url.click(Unexpected())
+            url.click(cast(Text, Unexpected()))
```
wsanchez: These casts are all because the tests are intentionally passing in incorrect types.

src/hyperlink/_url.py
wsanchez: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#type-annotations

wsanchez:
@mahmoud I know this is a honker diff… Since it's annotations and not "real code changes", I'll suggest that this is low-risk enough that if the typing is not perfect, we can address it with bug fixes; also, we aren't going to publish the type hints to clients right away, and I'll test them against Klein once we merge here before we do so.

src/hyperlink/_url.py
mahmoud: Yeah, subencoding is in the TODO as this... thing. So when bytes are percent encoded and stuck in the URL, theoretically they can be Latin-1 or another encoding.  Does that make sense?

My thought was that maybe the subencoding arg could bubble up to the URL API, as an escape valve for non-UTF8 subencodings.. If you have thoughts about that, I'd love to hear them. Maybe it's just adding a warning box somewhere in the hyperlink docs: "hyperlink only supports utf-8 encoded text within URL parameters, etc."

src/hyperlink/_url.py
mahmoud: This used to be my style for a long time (mostly due to a harsh-grading TA), then I found guard statements fit my brain better (and bought me back indentation levels). Would love to discuss the single-return practice sometime.

src/hyperlink/_url.py
wsanchez: OK that makes sense.  I'll assert that we cross that bridge when we get to it, but that it would be better to refactor this code when we do so, such that we aren't having `_percent_decode` with a polymorphic signature.

src/hyperlink/_url.py
```diff
@@ -1705,59 +1822,85 @@ def child(self, *segments):
...
+        ])
         return self._path
 
     @property
     def query(self):
+        # type: () -> QueryPairs
         try:
-            return self._query
+            return cast(
+                QueryPairs,
```
mahmoud: Dang, are these casts in the critical path logic necessary? On a previous readthrough I had only seen the ones in the test file. 

I remember some past PRs getting a slight grilling for the potential perf impact of converting between generator expressions and list comprehensions, is there somewhere I can learn about the tradeoffs of using this approach, or do you mind giving me a quick primer? Like, is mypy not able to infer these as `(str, obj)` or `(str, str/None)`?

mahmoud: Like, pretty sure pytype can do it, especially considering all the other annotations you've put in: https://github.com/google/pytype

Are we wedded to mypy as the checker, or would you be open to trying pytype (I've heard good things, admittedly, from googlers, so I'm cautious too)

mahmoud:
@wsanchez The tests against Klein sound like a great idea, thank you in advance for doing that. It's a lot of lines, but I'm not worried about the comment-style annotations. Per my other comments, my main concerns are about abandoning non-utf8 subencodings (which may be fine), and also the typing-related functions creeping into the primary code paths.

src/hyperlink/_url.py
wsanchez: `cast()` is supposed to be equivalent to `lambda x: x` at runtime, so there's a function call of overhead, and that's is.  But that's not zero.

I'm pretty annoyed that some of these casts are necessary to appease `mypy`… I think one option is that we can do more `# type: ignore` hooey here and wait on `mypy` to fix what I suspect are  some bugs instead of casting here.  This was an attempt to narrow down the ignores to a minimum, then see if `mypy` can incrementally address the issues.

src/hyperlink/_url.py
wsanchez: So the challenge here is that `_percent_decode` with `subencoding=False` returns `bytes` instead of `Text` (aka `unicode`).  That would require that we annotate it as `Union[Text, bytes]` and a large number of callers would then have to `cast` the return to `Text` (or `bytes` in the future).

Even in cases where you could previously specify `subencoding=False`, there are cases in this function where you get `Text` instead of `bytes`, which seems problematic.

So what I'd suggest is that when we want to get raw bytes out, that we add a new function that always returns bytes, possibly factoring out the middle bit (for for loop) form here.

src/hyperlink/_url.py
wsanchez: I don't object to either, but the code was doing both, so I figured it would be simpler to pick one or the other for consistency.  In this case I thought "figure out what family is" would be one step and "return a result with family and host" would follow.  Alternatively, we could have changed the `family = FOO` lines to `return FOO, host`.

src/hyperlink/_url.py
```diff
@@ -1705,86 +1821,111 @@ def child(self, *segments):
...
-        except AttributeError:
-            pass
-        self._path = tuple([_percent_decode(p, raise_subencoding_exc=True)
-                            for p in self._url.path])
+        # type: () -> Sequence[Text]
+        if not hasattr(self, "_path"):
+            self._path = tuple([
+                _percent_decode(p, raise_subencoding_exc=True)
+                for p in self._url.path
+            ])
```
wsanchez: Checking for the attribute instead of catching `AttributeError` helps `mypy` out.
I think it's a little more readable as well… (but only a little).

src/hyperlink/_url.py
wsanchez: OK I got rid of a bunch of calls to `cast()`

src/hyperlink/_url.py
wsanchez: ARGH! "Python 2/3 compatible annotations aren’t currently supported by Sphinx and won’t show up in the docs."

wsanchez:
@mahmoud I put type info back into the doc strings because the doc generator doesn't handle Python 2-compatible type comments.

I used `typing` style types, which I think are more accurate, if more verbose, and I only did this for public API.

We can remove them and switch to inline type comments whenever 2.7 support goes away. Then the doc generator will pick them up from the type hints.

src/hyperlink/_url.py
glyph: If my understanding is correct here we could do this with an `@override`?

wsanchez: Perhaps, but wouldn't it be safer to do the separate function? It's internal API, so type checking here is for us to catch our own mistakes.

mahmoud: Yeah, I'm fine handling this as a refactor later. We'll always have the git history :)

src/hyperlink/_url.py
mahmoud: Weird that mypy has trouble with that, given Python's classic argument about TOCTOU / EAFP. 

Another issue is that `hasattr()` returns False on non-AttributeErrors on py2. But realistically it's not a big regression, so it's not a blocker.

PR #84, opened 2019-11-06
Provide Hypothesis strategies

wsanchez:
Provide strategies for use with [Hypothesis](https://hypothesis.readthedocs.io/).

This brings over [the strategies from Klein](https://github.com/twisted/klein/blob/master/src/klein/test/_strategies.py), reformats code to match Hyperlink, add test cases, and fixes a few things.


mahmoud:
Wow! Nice work! I'm curious (and pretty green wrt hypothesis), is the convention for packages to have a `strategies.py` in them like that?

wsanchez:
@mahmoud: well, in Hypothesis itself, strategies are in `hypothesis.strategies`, which is where I got that.

I can see what other packages are doing.

src/hyperlink/strategies.py
```diff
@@ -0,0 +1,292 @@
+# -*- coding: utf-8 -*-
+"""
+Hypothesis strategies.
+"""
+
+try:
+    import hypothesis
+    del(hypothesis)
```
mahmoud: Is this some py4 forward compat?? jkjk, but `del` is still a statement in py3 :)

wsanchez:
@mahmoud: Here are some examples from other libraries [listed in the Hypothesis documentation](https://hypothesis.readthedocs.io/en/latest/strategies.html)…

Hypothesis itself:
- hypothesis: `from hypothesis.strategies import text`

These only provide strategies, as opposed to being included as part of some other library:
- hypothesis-fspaths: `from hypothesis_fspaths import fspaths`
- hypothesis-geojson: `from hypothesis_geojson import features`
- hs-dbus-signature: `from hs_dbus_signature import dbus_signatures`
- hypothesis-sqlalchemy: `from hypothesis_sqlalchemy import tabular`
- hypothesis-ros: `from hypothesis_ros import message_fields`
- hypothesis-csv: `from hypothesis_csv.strategies import csv`
- Hypothesis-networkx: `from hypothesis_networkx import graph_builder`
- Hypothesis-bio: `from hypothesis-bio.strategies import sequence`
- hypothesmith: `from hypothesmith import from_grammar`

I'm not sure how much precedent we can draw from the above, given that they aren't built into an existing library…

I would be fine with `hyperlink.hypothesis` instead of `hyperlink.strategies`.  It does sound a bit more specific, if alliterative.  :-)

src/hyperlink/strategies.py
wsanchez: Oh, I can't help but use parens, oops.

This was just to separate the "do we have hypothesis?" test from where we import what we need; I was finding it annoying to have one import line up by it's lonesome.  :-?

wsanchez: Perhaps you were asking why "do we have hypothesis?" is a question.
It's (perhaps weirdly) a Python 3.4 thing.
Hypothesis supports 2.7 and >=3.5, but not 3.0-3.4.

mahmoud: Heh, nope just eyeing the parens.

wsanchez:
@mahmoud Any preferences for the module name? `hyperlink.hypothesis`?

wsanchez:
@mahmoud @glyph I'd love to get this and the type hints into a release soon. :-)

glyph:
This feels very heavy-weight to me as a necessary runtime dependency, particularly given the chunky data-file. Is there maybe a good reason that these generally go into separate packages?  This is the sort of thing you would want as a test or dev dependency but not something installed with your app if you're at all resource constrained.

mahmoud:
That's true, though hyperlink isn't really the most performance-centric URL library on the block. I've always seen it as "good enough".

I don't think that changes with this PR. These new modules and assets aren't imported by default, right? Nor does it add more `install-requires`? If there's concern about the storage requirements of the 130kb data file, I bet it compresses way down. Not sure how often it needs to be updated.

wsanchez:
Right, the memory impact is low unless you actually use it. The IDNA table does add 135KB to the package… I can compress the file if we want.

wsanchez:
The IDNA file compressed is 26KB.

wsanchez:
Well 2.7 on Windows sucks.

mahmoud:
lol, spill the `t`

wsanchez:
I added the `t` because `csv.reader` pukes on `bytes` IO in Py3. I think a generator expression will work around this.

glyph:
> That's true, though hyperlink isn't really the most performance-centric URL library on the block. I've always seen it as "good enough".

we don't have performance budgets or anything, but doubling the required installed size of the library for a purely test-related dependency seems like something that at least bears consideration - especially given that the pattern seems to be to put these things in other libraries.  After compressing the data file, it's more like a 50% increase than 100%, but still, it's pretty significant.

I don't want to set hard perf budgets but I would like to be able to put hyperlink on a raspberry pi or router without really thinking too hard about it; there are lots of applications in edge networking with tiny devices that can really use robust URL parameterization.

I'm +1 on the functionality here, but still -0 on including it in the same package; it really just seems natural to have a separate hypothesis repository, as the 9 examples above do.

glyph:
Also, I'm kind of curious - where did this IDNA file even come from?  Are there any copyright concerns?

wsanchez:
OK, well if you really feel strongly about it, I can make another repo, but I think the site issue here is a minor factor, FWIW.

The file comes from IANA: https://www.iana.org/assignments/idna-tables-10.0.0/idna-tables-properties.csv

mahmoud:
I tend to agree with @wsanchez. But on the off chance we were to get some valid complaint we were overlooking, we could probably split it out into an optional requirement and use a namespace package to keep the imports the same. Would that make sense as a contingency plan?

glyph:
I'd rather go the other way - make a second top-level package first.  Namespace packages are a gigantic pain

glyph:
pull it in when we need it, that is

wsanchez:
I'm not a fan of pulling it down from the network in tests.

I'll make it a separate package.

mahmoud:
@wsanchez "it" in the previous comment referring to all the strategies? Personally I would've merged, but if you're already set on a separate package, should we close this PR and with a note about the package name / other next steps?

glyph:
> @wsanchez "it" in the previous comment referring to all the strategies? Personally I would've merged, but if you're already set on a separate package, should we close this PR and with a note about the package name / other next steps?

Let's do that.  I'll close now, we can attach the comment when @wsanchez creates the new package.

@wsanchez, for resiliency, would you mind adding us both on PyPI when you create the upload? in my ~ copious spare time ~ I'm trying to build automated PyPI release pipelines.

glyph:
(Leaving the branch around for now, but when this is moved please hit that "delete branch" button)

wsanchez:
I'd prefer to keep the PR open until then, I think, since this is the only thing tracking that. I can mark it draft.

wsanchez:
@mahmoud "it" meant the IDNA file that @glyph thinks is huge. I don't think that grabbing data from the Internet during tests is cool.

wsanchez:
I also would rather have just merged, but @glyph is only slightly crazy, so I'm gonna try to figure out how to add this as a second package in this repo.

mahmoud:
Oh, same repo, didn't catch that first time around. Will be cool to see what you come up with.

wsanchez:
I haven't really figured out a great way to have two projects in a repo.  I'm sure it's doable, but fighting with `distutils` isn't super entertaining.

So I'd like to to go back to plan A (merge this) and if someone really finds it annoying to have a slightly larger package, then perhaps that person would be more motivated to make it work.

glyph:
> So I'd like to to go back to plan A (merge this) and if someone really finds it annoying to have a slightly larger package, then perhaps that person would be more motivated to make it work.

Sounds good to me.

mahmoud:
OK codecov is fixed and CI is green, I'm calling it good and merging! Big thanks and congrats all around!

PR #85, opened 2019-11-08
Add Tox environment for building the documentation.

wsanchez:
`tox -e docs` will generate `.tox/docs/html`

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -69,7 +69,6 @@ A couple of notable helpers used by the :class:`~hyperlink.URL` type.
 
 .. autoclass:: hyperlink.URLParseError
 .. autofunction:: hyperlink.register_scheme
-.. autofunction:: hyperlink.parse_host
 .. autofunction:: hyperlink.parse
 
 .. TODO: run doctests in docs?
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -61,7 +61,6 @@ URLs have many parts, and URL objects have many attributes to represent them.
 .. autoattribute:: hyperlink.URL.userinfo
 .. autoattribute:: hyperlink.URL.user
 .. autoattribute:: hyperlink.URL.rooted
-.. autoattribute:: hyperlink.URL.family
 
 Low-level functions
 -------------------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -70,5 +70,6 @@ A couple of notable helpers used by the :class:`~hyperlink.URL` type.
 .. autoclass:: hyperlink.URLParseError
 .. autofunction:: hyperlink.register_scheme
 .. autofunction:: hyperlink.parse_host
+.. autofunction:: hyperlink.parse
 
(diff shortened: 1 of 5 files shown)


docs/api.rst
```diff
@@ -61,7 +61,6 @@ URLs have many parts, and URL objects have many attributes to represent them.
 .. autoattribute:: hyperlink.URL.userinfo
 .. autoattribute:: hyperlink.URL.user
 .. autoattribute:: hyperlink.URL.rooted
-.. autoattribute:: hyperlink.URL.family
```
wsanchez: Ain't such a thing as `URL.family`...

mahmoud:
I think `parse_host` is deprecated and you're right, `parse` should definitely definitely get documented!

docs/api.rst
```diff
@@ -71,5 +70,6 @@ A couple of notable helpers used by the :class:`~hyperlink.URL` type.
 .. autoclass:: hyperlink.URLParseError
 .. autofunction:: hyperlink.register_scheme
 .. autofunction:: hyperlink.parse_host
+.. autofunction:: hyperlink.parse
```
wsanchez: This one is public but not in the docs…

wsanchez:
> I think `parse_host` is deprecated and you're right, `parse` should definitely definitely get documented!

OK I took `parse_host` back out of `__init__.py` and I removed it from the docs.

We might want to consider just removing that code, if it's not available except via `hyperlink._url`, which we wouldn't want to encourage people to import.  That said, that's a different PR.

wsanchez:
@mahmoud This look OK?

mahmoud:
@wsanchez hey sorry about the wait, yes, this looks mostly OK. I think it's a bit on the bleeding edge there with using py3.8, but I assume you've rendered and it looks ok? 

I've got some weird habits like using [sphinx-autobuild](https://pypi.org/project/sphinx-autobuild/), which I'm hoping continue to work (says it supports py3.6). As for those, I'll just deal with those bridges when I cross em.

wsanchez:
I'll give `sphinx-autobuild` a spin.

wsanchez:
> @wsanchez hey sorry about the wait, yes, this looks mostly OK. I think it's a bit on the bleeding edge there with using py3.8, but I assume you've rendered and it looks ok?

Looking at the rendered result… looks the same as on `hyperlink.readthedocs.io`.

wsanchez:
> @wsanchez hey sorry about the wait, yes, this looks mostly OK. I think it's a bit on the bleeding edge there with using py3.8, but I assume you've rendered and it looks ok?
> 
> I've got some weird habits like using [sphinx-autobuild](https://pypi.org/project/sphinx-autobuild/), which I'm hoping continue to work (says it supports py3.6). As for those, I'll just deal with those bridges when I cross em.

Neat.  I added a `docs-auto` Tox environment, which will run `sphinx-autobuild` for you.

PR #86, opened 2019-11-08
Add flake8 and mypy to Travis

wsanchez:
Travis isn't running `flake8` or `mypy`.  That seems a bit silly.


wsanchez:
Waiting on #82 before proceeding.

wsanchez:
Should probably also add `docs`, after #85

wsanchez:
@glyph `flake8` is catching an unused variable in a new test you added.  Can you check on whether that's meant to be used to removed?

wsanchez:
@glyph: `attempted_rooted_replacement`

src/hyperlink/test/test_url.py
```diff
@@ -1112,7 +1112,6 @@ def test_autorooted(self):
 
         attempt_unrooted_absolute = URL(host="foo", path=['bar'], rooted=False)
         normal_absolute = URL(host="foo", path=["bar"])
-        attempted_rooted_replacement = normal_absolute.replace(rooted=True)
```
wsanchez: @glyph: here

glyph: weird. that's a real bug. we should be asserting something about that object

wsanchez: I think that asserting something about that object is unrelated to this PR, though, and is an example of why landing this PR would be good :-)

wsanchez: Not sure how you want to address that.  Can you open a PR with a fix or an issue to track it?

wsanchez: For reference, this came via #91

glyph: Yeah, I guess what I meant was *I* should have asserted something about that object :).

You can feel free to just clean it up here for the time being, and assign any resulting issue to me

wsanchez: Issue #99 for replacing this.

PR #87, opened 2019-11-08
MOAR BADGERS, er… badges

wsanchez:


wsanchez:
`requirements.io` needs registration

mahmoud:
Hehe, I kind of kept these intentionally minimal and evergreen. 

* Build: All too often I run across a project with some transient travis error that mars the README. Not sure I want to risk that accidentally being our initial impression, right?
* Coverage: Our coverage is actually pretty high and will only get higher, so why not flaunt it, sure.
* Requirements: Our install_requires is basically unpinned, so I dunno if this yells about requirements.txt or what, but another one that I'd argue is less of a selling point.
* PyPI python version compat: Sure! I think we're pretty forward compatible, so my only concern would be this falling out of date and generating concern that we don't support 3.9 or something.

wsanchez:
> * Coverage: Our coverage is actually pretty high and will only get higher, so why not flaunt it, sure.

I do like bragging.

> * Build: All too often I run across a project with some transient travis error that mars the README. Not sure I want to risk that accidentally being our initial impression, right?
> * Requirements: Our install_requires is basically unpinned, so I dunno if this yells about requirements.txt or what, but another one that I'd argue is less of a selling point.

These are more for the developers.  I go to GitHub pages for projects I work on fairly regularly, and I I see a build badge that's red, I know there's a thing to fix before I go work on some other thing.

The requirements badge will check `tox.ini`, and serves as a reminder to update the pinned dependencies in there.  We pin those for testing stability, but should update them periodically.  Requirements.io makes it really easy to identify which need updates.

> * PyPI python version compat: Sure! I think we're pretty forward compatible, so my only concern would be this falling out of date and generating concern that we don't support 3.9 or something.

This one is already on master. :-)

wsanchez:
> This one is already on master. :-)

No wait, untrue.  This uses setup.py IIRC.  Hold on.

wsanchez:
We have email for build badge, so I think nixing that is reasonable.  I'm a fan of requirements, tho…

wsanchez:
[Preview here](https://github.com/python-hyper/hyperlink/tree/badger-badger-badger)

wsanchez:
@glyph just provided an example: https://github.com/python-hyper/hyperlink/pull/88

PR #91, opened 2019-11-10
attempt at a fix for the rooted flag

### LINKED ISSUE #90: rooted flag exposes serialization ambiguity
This should not be possible:
```
>>> import hyperlink
>>> a = hyperlink.URL(path=['', 'foo'], rooted=False)
>>> b = hyperlink.URL(path=['foo'], rooted=True)
>>> a == b
False
>>> a
URL.from_text('/foo')
>>> b
URL.from_text('/foo')
>>> a.normalize() == b.normalize()
False
```
I think that if `path` ever starts with a `''` when `rooted=False`, we ought to flip the flag and remove the path segment.

glyph:
Fixes #90

### DIFF
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -761,7 +761,9 @@ class URL(object):
        query (tuple): The query parameters, as a dictionary or
           as an iterable of key-value pairs.
        fragment (unicode): The fragment part of the URL.
-       rooted (bool): Whether or not the path begins with a slash.
+       rooted (bool): A rooted URL is one which indicates an absolute path.
+          This is True on any URL that includes a host, or any relative URL
+          that starts with a slash.
        userinfo (unicode): The username or colon-separated
           username:password pair.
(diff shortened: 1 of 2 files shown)


glyph:
@wsanchez @mahmoud anybody up for a review

wsanchez:
This looks fine to me, given the way the code works now, but I guess I don't understand why `rooted` is an argument to `URL` in the first place.

Shouldn't it just be a computed (perhaps cached) property instead of a user-defined attribute?

wsanchez:
I guess my question is really just aspirational; the API is what it is.

glyph:
> I guess my question is really just aspirational; the API is what it is.

I think the question still deserves answering, since this API is *super* subtle.  The main thing I'd say about this is that it does make a lot more sense than attempting to deal with nonsensical values for `path`, but still always get the right number of slashes, and not use different data structures for relative and absolute URLs.

PR #102, opened 2019-12-29
Update supported Python versions

hugovk:
* Update Trove classifiers for supported Python versions: add 3.7 and 3.8
* Update docs for tested Python versions: add 3.7 and 3.8
* Add `python_requires` to `setup.py` to help pip install the best version of this library for the user's running Python


setup.py
```diff
@@ -34,16 +34,21 @@
...
+      python_requires='>=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
       classifiers=[
           'Topic :: Utilities',
           'Intended Audience :: Developers',
           'Topic :: Software Development :: Libraries',
           'Development Status :: 5 - Production/Stable',
+          'Programming Language :: Python :: 2',
           'Programming Language :: Python :: 2.6',
           'Programming Language :: Python :: 2.7',
+          'Programming Language :: Python :: 3',
```
wsanchez: Are these two additions appropriate when we list specific sub-versions that are supported?
If yes, should we add `Programming Language :: Python`?

hugovk: Yes, it's common to list those as well. Sometimes `Programming Language :: Python` is there too.

For example, see pip:

* https://github.com/pypa/pip/blob/master/setup.py

Shall I add that one also?

wsanchez: Sorry, I forgot this was still open.

It's hard to argue with the `pip` team about packaging, so following their lead seems like a good idea.

Issue #24, opened 2017-07-03
remove "uses_netloc" parameter and attribute to .__init__ and .replace


glyph:
In twisted.python.url I tried *very* hard to eliminate the use of the confusing, vague, and antiquated term "netloc".  (NB: neither https://tools.ietf.org/html/rfc3986 nor https://url.spec.whatwg.org includes the string "netloc").  I would therefore like to eliminate its use in `uses_netloc` as well.

`uses_netloc` describes an attribute of a scheme.  However, in its current incarnation it can be made nonsensically inconsistent with the scheme it's actually using.  Additionally, schemes also have other attributes, such as a name, and a default port number.  (Possibly more in the future?)  As such, the presence of the scheme registry makes URL objects kinda/sorta mutable after the fact.

My suggestion for a replacement would be for `URL` to reference a structured `scheme` *object*; the external scheme registry could then be a collection of such objects.  If a URL is created before the registry is populated, its scheme object could then be inconsistent with the global one, but it could be replaced.  At this point, I believe schemes would just have the three attributes given above (i.e. the parameters to `registerScheme`.

mahmoud:
WHATWG and 3986 love to avoid talking about the slashes, but I don't know what to call them other than "netloc slashes", and `uses_netloc_slashes` seems a bit much.

Naming aside, `uses_netloc` is there for roundtripping and "just works"-ness. While I tried *very hard* to create a best-in-class scheme registry, with ~60 preregistered schemes, I still see it as inevitable that someone will have a scheme that's _not_registered. (`postgres` and other SQLAlchemy conventional driver names are such examples.)

In those cases, as mentioned in #23, `URL` uses `uses_netloc` to pass forward whether the original URL passed to `from_text()` used the netloc slashes. Not only is explicit better than implicit, but it's real information that we would be dropping on the ground otherwise.

If we switched to an object-based `scheme`, we could generate Scheme objects on the fly for unrecognized schemes, but there's nothing stopping people from basically doing the same thing. It'd be slightly harder, but I'd contend that it's also currently quite hard to make this mistake.

glyph:
WHATWG doesn't care much about the slashes since they're just "the things that come after http://" :) but 3986 specifically documents them; they are the slashes that come before the authority.

[...]

PR #112, opened 2020-02-16
fix up inconsistencies in parsing & textual representation of 'rooted…

### LINKED ISSUE #111: Rooted flag still causing serialization oddities
I thought this was the same bug as #90, but @glyph's #91 didn't fix it, so I guess it's at least slightly distinct.

```
_________ TestURL.test_reproduce_my_rooted_oddity _________

self = <hyperlink.test.test_url.TestURL testMethod=test_reproduce_my_rooted_oddity>

    def test_reproduce_my_rooted_oddity(self):
        a = URL(scheme='udp', port=4900)
        b = URL.from_text('udp://:4900')
        assert str(a) == str(b)
        assert a.asText() == b.asText()
>       assert a == b
E       AssertionError: assert URL.from_text('udp://:4900') == URL.from_text('udp://:4900')
E         -URL.from_text('udp://:4900')
E         +URL.from_text('udp://:4900')

src/hyperlink/test/test_url.py:1101: AssertionError
```

That assertion error is *entertainingly* befuddling 😃 

After a cursory inspection of `a` and `b` there, the only differences I found in the instance `__dict__` were `_rooted` and `_uses_netloc`.

The same failure happens with `URL(scheme='udp', host='', port=4900)`, but doing `URL(scheme='udp', port=4900, rooted=True)` (with or without `host`) makes the test pass.

(this was tested on 688233a2fccc470f8c822ccf91ea53ec03d1e427)

glyph:
…' and 'uses_netloc'

Fixes #111 

In addition to closing a few more holes with `rooted`, this addresses a bunch of weirdnesses with `uses_netloc`, trying to close the gap between the way the URL is serialized and the way that it is parsed; i.e. making it so that there is no "hidden state", and these flags are always consistent with the other parts of the URL that they influence.

### DIFF
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -815,9 +815,18 @@ class URL(object):
           that starts with a slash.
        userinfo (Text): The username or colon-separated
           username:password pair.
-       uses_netloc (bool): Indicates whether ``://`` will appear to separate
-           the scheme from the path, even in cases where no host is present.
-           May be implied by scheme, or set explictly.
+       uses_netloc (Optional[bool]): Indicates whether ``://`` (the "netloc
+           separator") will appear to separate the scheme from the *path* in
+           cases where no host is present.  Setting this to ``True`` is a
+           non-spec-compliant affordance for the common practice of having URIs
+           that are *not* URLs (cannot have a 'host' part) but nevertheless use
+           the common ``://`` idiom that most people associate with URLs;
+           e.g. ``message:`` URIs like ``message://message-id`` being
+           equivalent to ``message:message-id``.  This may be inferred based on
+           the scheme depending on whether :func:`register_scheme` has been
+           used to register the scheme and should not be passed directly unless
+           you know the scheme works like this and you know it has not been
+           registered.
 
     All of these parts are also exposed as read-only attributes of
     URL instances, along with several useful methods.
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -1028,8 +1028,8 @@ class URL(object):
     def uses_netloc(self):
         # type: () -> Optional[bool]
         """
-        Whether the textual URL representation will contain a ``://`` netloc
-        separator.
+        Indicates whether ``://`` (the "netloc separator") will appear to
+        separate the scheme from the *path* in cases where no host is present.
         """
         return self._uses_netloc
 
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -1158,8 +1158,19 @@ class URL(object):
                 slash.
             userinfo (Text): The username or colon-separated username:password
                 pair.
-            uses_netloc (bool): Indicates whether rooted paths should include a
-                scheme-separator by default.
+            uses_netloc (bool): Indicates whether ``://`` (the "netloc
+                separator") will appear to separate the scheme from the *path*
+                in cases where no host is present.  Setting this to ``True`` is
+                a non-spec-compliant affordance for the common practice of
+                having URIs that are *not* URLs (cannot have a 'host' part) but
+                nevertheless use the common ``://`` idiom that most people
+                associate with URLs; e.g. ``message:`` URIs like
+                ``message://message-id`` being equivalent to
+                ``message:message-id``.  This may be inferred based on the
+                scheme depending on whether :func:`register_scheme` has been
+                used to register the scheme and should not be passed directly
+                unless you know the scheme works like this and you know it has
+                not been registered.
 
         Returns:
             URL: A copy of the current :class:`URL`, with new values for
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -815,9 +815,9 @@ class URL(object):
           that starts with a slash.
        userinfo (Text): The username or colon-separated
           username:password pair.
-       uses_netloc (bool): Indicates whether two slashes appear
-          between the scheme and the host (``http://eg.com`` vs
-          ``mailto:e@g.com``). Set automatically based on scheme.
+       uses_netloc (bool): Indicates whether ``://`` will appear to separate
+           the scheme from the path, even in cases where no host is present.
+           May be implied by scheme, or set explictly.
 
     All of these parts are also exposed as read-only attributes of
     URL instances, along with several useful methods.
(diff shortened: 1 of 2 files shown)


glyph:
cc @cdunklau if you want to review this to see if it fully addresses your issue.

cdunklau:
> cc @cdunklau if you want to review this to see if it fully addresses your issue.

Yep, I can't reproduce oddities with this branch.

src/hyperlink/test/test_url.py
```diff
@@ -1116,6 +1138,18 @@ def test_autorooted(self):
         self.assertEqual(normal_absolute.rooted, True)
         self.assertEqual(attempt_unrooted_absolute.rooted, True)
 
+    def test_rooted_with_empty_non_none_host(self):
+        # type: () -> None
+        """
+        The C{rooted} constructor argument will be ignored on URLs that include
+        a scheme.
```
cdunklau: This docstring doesn't seem to match with the test, as the ctor arg isn't involved. Maybe test that all four of these match `URL.from_text('udp://:4900')`? :

```
URL(scheme='udp', port=4900)
URL(scheme='udp', port=4900, rooted=False)
URL(scheme='udp', host='', port=4900)
URL(scheme='udp', host='', port=4900, rooted=False)
```

glyph: Updated to more closely match the desired behavior that is being tested.

glyph: Thanks for the review!

src/hyperlink/_url.py
```diff
@@ -815,9 +815,9 @@ class URL(object):
           that starts with a slash.
        userinfo (Text): The username or colon-separated
           username:password pair.
-       uses_netloc (bool): Indicates whether two slashes appear
-          between the scheme and the host (``http://eg.com`` vs
-          ``mailto:e@g.com``). Set automatically based on scheme.
+       uses_netloc (bool): Indicates whether ``://`` will appear to separate
+           the scheme from the path, even in cases where no host is present.
+           May be implied by scheme, or set explictly.
```
wsanchez: Should this say that one probably should only set it explicitly if you've registered a scheme not known to hyperlink?

src/hyperlink/_url.py
```diff
@@ -882,15 +882,28 @@ def __init__(
...
-        if (not self._rooted) and self._path and self._path[0] == '':
+            self._uses_netloc = True
+        if (not self._rooted) and self.path[:1] == (u'',):
             self._rooted = True
             self._path = self._path[1:]
+        if not will_have_authority and self._path and not self._rooted:
+            # If, after fixing up the path, there *is* a path and it *isn't*
+            # rooted, then we are definitely not using a netloc; if we did, it
+            # would make the path (erroneously) look like a hostname.
+            self._uses_netloc = False
```
wsanchez: [This conversation](https://github.com/python-hyper/hyperlink/issues/24#issuecomment-312569915) makes me think that if `uses_netloc ` is not `None`, you should be comparing `self._uses_netloc` against `uses_netloc` and raising an exception if they do not match.  Same for `self._rooted`.

src/hyperlink/_url.py
```diff
@@ -1006,6 +1019,8 @@ def userinfo(self):
     def uses_netloc(self):
         # type: () -> Optional[bool]
         """
+        Whether the textual URL representation will contain a ``://`` netloc
+        separator.
```
wsanchez: This should match the first sentence from `__init__`.

src/hyperlink/_url.py
```diff
@@ -1134,14 +1149,17 @@ def replace(
                 slash.
             userinfo (Text): The username or colon-separated username:password
                 pair.
-            uses_netloc (bool): Indicates whether two slashes appear between
-                the scheme and the host
-                (``http://eg.com`` vs ``mailto:e@g.com``)
+            uses_netloc (bool): Indicates whether rooted paths should include a
+                scheme-separator by default.
```
wsanchez: This should match `__init__`.

src/hyperlink/test/test_url.py
```diff
@@ -938,15 +950,25 @@ def test_netloc(self):
         # type: () -> None
         url = URL(scheme='https')
         self.assertEqual(url.uses_netloc, True)
+        self.assertEqual(url.to_text(), u'https://')
+        self.assertEqual(URL.from_text('https:').uses_netloc, False)
+        self.assertEqual(URL.from_text('https://').uses_netloc, True)
```
wsanchez: Should you throw in `https:/` also?

src/hyperlink/test/test_url.py
```diff
@@ -1116,6 +1138,31 @@ def test_autorooted(self):
         self.assertEqual(normal_absolute.rooted, True)
         self.assertEqual(attempt_unrooted_absolute.rooted, True)
 
+    def test_rooted_with_port_but_no_host(self):
+        # type: () -> None
+        """
+        URLs which include a netloc separator are inherently rooted, regardless
+        of whether they include one because they specify an explicit host or
```
wsanchez: "which include a netloc" followed by "whether they include one" is confusing me. I think you mean "regardless of how they came to include one…"

mahmoud:
@glyph if you've got a little time to address Wilfredo's comments, happy to review and merge this.

src/hyperlink/_url.py
glyph: I don't entirely disagree, but if we do that I'd like to keep that in a separate change, because that pushes us quite far down the direction of "`uses_netloc` is a bad idea and should not have been added", which I am increasingly believing is actually what we should be doing with this value.  Rather than raising in specific circumstances which you can't know about in advance because you don't and can't know what other code has done global cache-mangling with `register_scheme`.  So I there's a lot to discuss here beyond enforcing correctness, which is what this branch is trying to do.

Mostly folks just shouldn't be using this feature though.

glyph:
@wsanchez @mahmoud I think that's everything.  Whoever has time...

src/hyperlink/_url.py
```diff
@@ -815,9 +815,18 @@ class URL(object):
           that starts with a slash.
        userinfo (Text): The username or colon-separated
           username:password pair.
-       uses_netloc (bool): Indicates whether two slashes appear
-          between the scheme and the host (``http://eg.com`` vs
-          ``mailto:e@g.com``). Set automatically based on scheme.
+       uses_netloc (Optional[bool]): Indicates whether ``://`` (the "netloc
+           separator") will appear to separate the scheme from the *path* in
```
mahmoud: Because `:` is always present when there's a scheme, wouldn't `//` be the "netloc separator"?

glyph: I think I'd still go with `://` here, because while you're right about the colon, `//` can appear lots of other places in the URL without being a netloc separator, but an unescaped `://` always means "netloc separator"

mahmoud: Sure, if you feel like that's clearer.

PR #121, opened 2020-04-02
A few more more Tox cleanup bits

wsanchez:
While merging #106 forward, I found a few more bits to clean up.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,10 +2,10 @@
 
 envlist =
     flake8, mypy
-    test-py{26,27,34,35,36,37,38,py,py3}
+    test-py{26,27,34,35,36,37,38,py2,py3}
     coverage_report
-    packaging
     docs
+    packaging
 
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -265,7 +265,7 @@ commands =
     sphinx-build                           \
         -b html -d "{envtmpdir}/doctrees"  \
         "{toxinidir}/docs"                 \
-        "{toxworkdir}/docs/html"
```
wsanchez, line 268: This is to match other projects, and similar to `/htmlcov/`.

mahmoud: Just curious, other hyper projects, or which other projects?

wsanchez: Klein in particular; I'm trying to keep the configs relatively similar for sanity.

glyph: /me whispers "we need a tool to keep them in sync"

wsanchez: I KNOW

PR #122, opened 2020-04-02
Note changes since 19.0.0

wsanchez:


mahmoud:
That's a good point, don't we need a changelog entry for #112?

wsanchez:
> That's a good point, don't we need a changelog entry for #112?

Perhaps… @glyph can write up a blurb for that…?

glyph:
"""
"""
Several bugs related to hidden state were fixed, making it so that all data on a URL object (including `rooted` and `uses_netloc`) is reflected by and consistent with its textual representation. This does mean that sometimes these constructor arguments are ignored, if it would create invalid or unparseable URL text.
"""

How's that?

glyph:
@wsanchez @mahmoud ⬆️

wsanchez:
OK, with Glyph's text added, this should be ready to do, as long as any additional merges add entries as appropriate.

PR #123, opened 2020-04-05
[requires.io] dependency update on master branch

glyph:


glyph:
@mahmoud I can't review this because "I" submitted it, so please rubber stamp

wsanchez:
This includes stuff I'm not sure it should.

tox.ini
```diff
@@ -187,7 +187,7 @@ basepython = {[default]basepython}
 skip_install = True
 
 deps =
-    coverage==4.5.4  # rq.filter: <5  # coverage 5.0 drops Python 3.4 support
```
wsanchez, line 190: rq.filter doesn't seem to be working

glyph: You have to commit it to `master` before it takes effect.  Committing to this branch doesn't do anything.

wsanchez:
@mahmoud or @glyph If you can thumbs up the docs change, I'll fix the `coverage` version up prior to merging (`requires.io` will keep stomping the fix in the meantime).

tox.ini
wsanchez: [It's on master](https://github.com/python-hyper/hyperlink/blob/master/tox.ini#L190)…

wsanchez: …12 days ago. Lemme reverting that again.

mahmoud:
So this is the only remaining PR, and I'm assuming it's a release blocker. Sound about right? Makes sense to me we'd want to bump these before 20.x?

wsanchez:
I think it's more of a nice-to-have than a blocker, since no functionality changes here.

The rendered documentation does change, and the types are more accurate, but it's not "prettier" and I haven't had time to figure out how to convince Sphinx to replace `hyperlink._url.URL` with `URL`, for example.

There's also CI not passing, I know one problem is `requires.io` updating things we don't want to update, but there may be others.

All that said, I'd be good with shipping from master.

wsanchez:
This PR is getting pretty wooly due to the Sphinx update… how about I break it into separate PR steps:
 1. Update requirements, and restrict Sphinx to <3.
 2. Update Sphinx and fix typing
 3. Zero this PR back out.

glyph:
Please do!

tox.ini
wsanchez: I can't figure out what we need to do to tell `requires.io` to stop updating this.  I must have the syntax wrong or something…  @glyph ?

glyph: I don't see comments-in-the-comment in any other projects I've done with requires.io, so maybe rather than:

```
    coverage==4.5.4  # rq.filter: <5  # coverage 5.0 drops Python 3.4 support
```
do
```
    # coverage 5.0 drops Python 3.4 support
    coverage==4.5.4  # rq.filter: <5
```

And then close this PR to see what it does when it reopens it.

wsanchez:
OK looks like this is all good now. Merging.

Issue #125, opened 2020-04-05
Document DecodedURL


twm:
There don't seem to be API docs for DecodedURL. As far as I can see it's only mentioned in the docstring of [hyperlink.parse](https://hyperlink.readthedocs.io/en/latest/api.html#hyperlink.parse) (which also mentions `EncodedURL` without explaining that `URL` _is_ `EncodedURL`).

URL's escaping behavior is inconsistent between path manipulation functions and querystring manipulators. The former escape, while the latter validate input:

```
Python 3.7.7 (default, Mar 10 2020, 15:16:38) 
>>> from hyperlink import URL, DecodedURL
>>> u = URL.from_text('https://example.com')
>>> u.child('foo/bar')
URL.from_text('https://example.com/foo%2Fbar')
>>> u.add('foo', '&')
Traceback (most recent call last):
  ...
ValueError: one or more reserved delimiters &# present in query parameter value: '&'
```

The documentation of `add()`, `set()`, etc. should at least mention this validation.

`DecodedURL` has consistent behavior: both escape as required:

```
>>> du = DecodedURL.from_text('https://example.com')
>>> du.child('foo/bar')
DecodedURL(url=URL.from_text('https://example.com/foo%2Fbar'))
>>> du.add('foo', '&')
DecodedURL(url=URL.from_text('https://example.com/?foo=%26'))
```

I find this behavior less surprising.

Issue #11, opened 2017-06-19
what if my password has a reserved delimiter in it?


glyph:
Right now if I do this:

```python
>>> from hyperlink import URL
>>> example = URL.fromText("https://example.com/")
>>> example.replace(userinfo="alpha:my#password")
```

I get this:

```python
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    example.replace(userinfo="alpha:my#password")
  File "/Users/glyph/.virtualenvs/tmp-a0b3197f7a1da77/lib/python3.6/site-package
s/hyperlink/_url.py", line 861, in replace
    userinfo=_optional(userinfo, self.userinfo),
  File "/Users/glyph/.virtualenvs/tmp-a0b3197f7a1da77/lib/python3.6/site-package
s/hyperlink/_url.py", line 615, in __init__
    self._userinfo = _textcheck("userinfo", userinfo, '/?#@')
  File "/Users/glyph/.virtualenvs/tmp-a0b3197f7a1da77/lib/python3.6/site-package
s/hyperlink/_url.py", line 410, in _textcheck
    % (''.join(delims), name, value))
ValueError: one or more reserved delimiters /?#@ present in userinfo: 'alpha:my#
password'
```

The API for setting a secret is sort of silly already (manually sticking a colon into the userinfo string), so "fixing" this might not involve any behavior change, but rather add a new argument (`.replace(secret=...?)`) but there should be some way to take a string that a user typed into a password box and embed it into the URL somehow without forcing the caller to do any wacky percent encoding of their own.

mahmoud:
Agreed, `userinfo` is a source of awkwardness. Could also do a helper function, `format_userinfo()` or something, as using passwords in the userinfo field is [quite strongly deprecated](https://tools.ietf.org/html/rfc3986#section-3.2.1).

glyph:
Self-authenticating URLs are a valuable construct.  There are good reasons to deprecate authentication-bearing URLs in a variety of contexts (like: don't let users click on them, maybe), but this deprecation has no teeth, specifically because there's no other way to address this use case.

My personal use-case for userinfo has to do with embedding database passwords into `postgresql://`-scheme URLs to feed to sqlalchemy, not clicking on in a web browser :).

mahmoud:
[...]

PR #126, opened 2020-04-10
Adding parse and DecodedURL docs

mahmoud:
#125 correctly points out, `hyperlink.parse()` and `hyperlink.DecodedURL()` aren't documented, so I put together something quick. 

I think we maybe talked about having URL _become_ DecodedURL at some point, but I'm not sure it's worth waiting for that. 

I think this PR basically brings the docs in line with recommended practices for using hyperlink.

Questions reviewers in the comments.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,11 +5,43 @@ Hyperlink API
 
 .. automodule:: hyperlink._url
 
+.. contents::
+   :local:
+
 Creation
 --------
 
-Before you can work with URLs, you must create URLs. There are two
-ways to create URLs, from parts and from text.
+Before you can work with URLs, you must create URLs.
+
+Parsing Text
+^^^^^^^^^^^^
+
+If you already have a textual URL, the easiest way to get URL objects
+is with the :func:`parse()` function:
+
+.. autofunction:: hyperlink.parse
+
+By default, :func:`~hyperlink.parse()` returns an instance of
+:class:`DecodedURL`, a URL type that handles all encoding for you, by
+wrapping the lower-level :class:`URL`.
+
+DecodedURL
+^^^^^^^^^^
+
+.. autoclass:: hyperlink.DecodedURL
+.. automethod:: hyperlink.DecodedURL.from_text
+
+The Encoded URL
+^^^^^^^^^^^^^^^
+
+The lower-level :class:`URL` looks very similar to the
+:class:`DecodedURL`, but does not handle all encoding cases for
+you. Use with caution.
+
+.. note::
+
+   :class:`URL` is also available as an alias,
+   ``hyperlink.EncodedURL`` for more explicit usage.
 
 .. autoclass:: hyperlink.URL
 .. automethod:: hyperlink.URL.from_text
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -39,9 +39,9 @@ library. The easiest way to install is with pip::
 
 Then, URLs are just an import away::
 
-  from hyperlink import URL
+  import hyperlink
 
-  url = URL.from_text(u'http://github.com/python-hyper/hyperlink?utm_source=readthedocs')
+  url = hyperlink.parse(u'http://github.com/python-hyper/hyperlink?utm_source=readthedocs')
 
   better_url = url.replace(scheme=u'https', port=443)
   org_url = better_url.click(u'.')
@@ -49,7 +49,7 @@ Then, URLs are just an import away::
   print(org_url.to_text())
   # prints: https://github.com/python-hyper/
 
-  print(better_url.get(u'utm_source'))
+  print(better_url.get(u'utm_source')[0])
   # prints: readthedocs
 
 See :ref:`the API docs <hyperlink_api>` for more usage examples.
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -3,16 +3,18 @@ u"""Hyperlink provides Pythonic URL parsing, construction, and rendering.
 
 Usage is straightforward::
 
-   >>> from hyperlink import URL
-   >>> url = URL.from_text(u'http://github.com/mahmoud/hyperlink?utm_source=docs')
+   >>> import hyperlink
+   >>> url = hyperlink.parse(u'http://github.com/mahmoud/hyperlink?utm_source=docs')
    >>> url.host
    u'github.com'
    >>> secure_url = url.replace(scheme=u'https')
    >>> secure_url.get('utm_source')[0]
    u'docs'
 
-As seen here, the API revolves around the lightweight and immutable
-:class:`URL` type, documented below.
+Hyperlink's API centers on the :class:`DecodedURL` type, which wraps
+the lower-level :class:`URL`, both of which can be returned by the
+:func:`parse()` convenience function.
+
 """  # noqa: E501
 
 import re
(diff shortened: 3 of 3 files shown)


src/hyperlink/_url.py
```diff
@@ -1763,8 +1775,18 @@ class DecodedURL(object):
...
             check for validity. Defaults to False.
 
+    .. note::
+
+      The :class:`DecodedURL` initializer takes a :class:`URL` object,
+      not URL components, like :class:`URL`. To programmatically
+      construct a :class:`DecodedURL`, you can use this pattern:
+
+        >>>  DecodedURL().replace(host='pypi.org', path=('projects', 'hyperlink').to_text()
+        "http://pypi.org/projects/hyperlink"
```
mahmoud, line 1785: Any opinion on adding a `DecodedURL.build(...)` or is this a fine pattern?

docs/api.rst
```diff
@@ -5,11 +5,43 @@ Hyperlink API
...
+:class:`DecodedURL`, but does not handle all encoding cases for
+you. Use with caution.
+
+.. note::
+
+   :class:`URL` is also available as an alias,
+   ``hyperlink.EncodedURL`` for more explicit usage.
 
 .. autoclass:: hyperlink.URL
 .. automethod:: hyperlink.URL.from_text
```
mahmoud, line 47: What do people think about switching over all "automethods" below this point to be DecodedURL methods?

For my money, I think it makes sense to have only one set of the URL methods given that the names are all the same, and there's only one behaviorial difference that they all share. Since DecodedURL is the recommended API, shouldn't all the methods point at that? Maybe for another PR though.

mahmoud:
@twm if you've got a sec, please take a look and let me know what you think. thanks!

docs/api.rst
wsanchez: I think it would be helpful to define an ABC which has the shared API and the docs can go there.

Doing so would help us catch any inconsistencies in that shared API between the two classes, and it would allow one to specify a type hint for an object that can be either, etc.

mahmoud: I agree an ABC could be useful for some. We'd have to call it AbstractUR, maybe BaseURL, or something, because I don't want to have to explain yet another URL type in the docs.

If everyone's ok with that direction, I would propose the ABC for a separate PR, and we can update the docs the rest of the way then.

wsanchez: Yup to `AbstractURL`, since `URL` is taken.

src/hyperlink/_url.py
```diff
@@ -1763,8 +1776,18 @@ class DecodedURL(object):
...
+    .. note::
+
+      The :class:`DecodedURL` initializer takes a :class:`URL` object,
+      not URL components, like :class:`URL`. To programmatically
+      construct a :class:`DecodedURL`, you can use this pattern:
+
+        >>> DecodedURL().replace(host='pypi.org', path=('projects', 'hyperlink').to_text()
+        "http://pypi.org/projects/hyperlink"
+
+    *(New in 18.0.0)*
```
twm, line 1789: [``.. versionadded::``](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded)?

twm:
I should say more clearly: the documentation changes here are exactly what I was looking for. Thank you!

src/hyperlink/_url.py
mahmoud: lol I knew I was forgetting something, lemme fix that

mahmoud: fixed, thanks!

mahmoud:
OK I forked off the AbstractURL discussion to #127. With that taken care of, I think these docs answer far more questions than they raise, so if there are no critical blockers, I'd like to get them deployed sooner rather than later. Any approvers?

docs/api.rst
mahmoud: Resolving by way of opening #127 :)

src/hyperlink/_url.py
```diff
@@ -1991,9 +2005,20 @@ class DecodedURL(object):
...
+
+        >>> print(DecodedURL().replace(scheme=u'https',
+        ... host=u'pypi.org', path=(u'projects', u'hyperlink')).to_text())
+        https://pypi.org/projects/hyperlink
+
+    .. versionadded:: 18.0.0
     """
 
-    def __init__(self, url, lazy=False):
+    def __init__(self, url=_EMPTY_URL, lazy=False):
```
wsanchez, line 2021: why the default?

docs/index.rst
```diff
@@ -39,17 +39,17 @@ library. The easiest way to install is with pip::
 
 Then, URLs are just an import away::
 
-  from hyperlink import URL
+  import hyperlink
 
-  url = URL.from_text(u'http://github.com/python-hyper/hyperlink?utm_source=readthedocs')
+  url = hyperlink.parse(u'http://github.com/python-hyper/hyperlink?utm_source=readthedocs')
```
wsanchez, line 44: FWIW I'm not a fan of `parse`, which can return two types, and would recommend that one use `URL.from_text` or `DecodedURL.from_text` directly instead, so this change seems like a minus to me, but I might be alone there.

src/hyperlink/_url.py
wsanchez, line 2021: why the default? Perhaps merits discussion outside of doc changes.

mahmoud: Sure, a couple reasons:

1. It's more parallel with `URL()`, which can be instantiated without arguments to get an empty URL.
2. This way to programmatically construct a `DecodedURL`, you don't need to also import `URL`. You can do `DecodedURL().replace(...)`.

docs/index.rst
mahmoud: Yeah I think `parse()` was added to shift recommended practice toward DecodedURLs, because they're safer. `parse()` returns a `DecodedURL` by default. I think it also looks pretty nice.

src/hyperlink/_url.py
mahmoud: Explained in the other discussion, here: https://github.com/python-hyper/hyperlink/pull/126#discussion_r409186280

mahmoud: Ah, and the reason for including it in these changes was because I really wanted to avoid people programmatically constructing the newly-exposed DecodedURL by doing

```
from hyperlink import URL, DecodedURL
DecodedURL(URL(...))
```
Without realizing that URL's initializer arguments aren't as rigorously decoded/encoded as the rest of DecodedURL's API.

I proposed adding a `.build()` or [`.from_parts()`](https://github.com/python-hyper/hyperlink/issues/11#issuecomment-370124799) if we want to shift to a better pattern. For now it's either that or [`parse('').replace(...)`](https://github.com/python-hyper/hyperlink/issues/11#issuecomment-370096947). I figured a default arg at least lets you explicitly use the type without an extra `parse()`/`from_text()` step.

twm: `from_parts()` sounds like a good addition. I have written `DecodedURL(URL())` &mdash; it's a bit awkward. Perhaps I am worrying too much (this is Python, everything is slow...) but I tend to look for APIs that avoid temporaries.

mahmoud:
codecov is having some system issues right now, I can't get codecov to come back with the coverage. [The report here](https://codecov.io/gh/python-hyper/hyperlink/pull/126/commits) is saying it's gone up and is around 98.56%. Going to admin merge on that basis.

src/hyperlink/_url.py
wsanchez: OK makes sense, and yes to `from_parts`.

PR #128, opened 2020-04-15
Clean up the Tox config a bit more

wsanchez:
- Move some settings from `[default]` to `[testenv]`, since they are only used there.
- There is no need to prefix testing config with `test:` any more, since `[testenv]` only runs tests.
- Add comments, because yay comments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,12 +17,6 @@ basepython = python3.8
 deps =
     idna==2.9
 
-    test: typing==3.7.4.1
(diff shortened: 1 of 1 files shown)


mahmoud:
codecov's still doing that thing where it's not reporting back. I tried refreshing the perms from github's end, resetting from the codecov side, and also installing codecov's github app. No progress yet, but maybe next PR. For now, admin-merging.

PR #130, opened 2020-06-23
Update requirements, cap Sphinx

wsanchez:


tox.ini
```diff
@@ -94,7 +95,7 @@ setenv =
     BLACK_LINT_ARGS=--check
 
 commands =
-    black {env:BLACK_LINT_ARGS:} setup.py src
+    black {env:BLACK_LINT_ARGS:} {posargs:setup.py src}
```
wsanchez, line 98: This is to allow one to specify specific files.

setup.py
```diff
@@ -27,11 +27,11 @@
     url=__url__,
     packages=find_packages(where="src"),
     package_dir={"": "src"},
-    package_data=dict(hyperlink=["py.typed",],),
+    package_data=dict(hyperlink=["py.typed"]),
     zip_safe=False,
     license=__license__,
     platforms="any",
-    install_requires=["idna>=2.5", 'typing ; python_version<"3.5"',],
+    install_requires=["idna>=2.5", 'typing ; python_version<"3.5"'],
```
wsanchez, line 34: Because we're adding `setup.py` to the files Flake8 looks at.

mahmoud: Appreciate the explanation :)

tox.ini
mahmoud: makes sense

PR #131, opened 2020-06-25
Update to Sphinx3

wsanchez:
This updates Sphinx to version 3 (from version 2).

One thing that is different is that Sphinx 3 will add type hint information to API documentation, for example:

```
hyperlink.parse(url: str, decoded: bool = True, lazy: bool = False) → Union[hyperlink._url.URL, hyperlink._url.DecodedURL]
```

This means that the manually added type information in the docstrings is no longer needed.

### DIFF
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -478,12 +478,12 @@ def register_scheme(text, uses_netloc=True, default_port=None):
     `file an issue`_!
 
     Args:
-        text (Text): A string representation of the scheme.
+        text: A string representation of the scheme.
             (the 'http' in 'http://hatnote.com')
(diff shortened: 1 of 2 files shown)


wsanchez:
Hrm.  I tried switching back to `Sphinx==2.4.4` and I still see the automatic type information in the generated API docs… so perhaps I'm wrong and that wasn't a Sphinx 3 feature…?

In which case, all this PR is really doing is updating Sphinx and removing the duplicate information.

wsanchez:
One this that I would like to fix is `hyperlink._url.URL` should just be `hyperlink.URL` or even `URL`, but I don't know how to do that.

glyph:
(We can figure out the class-name replacement later)

PR #134, opened 2020-08-28
Install IDNA tables file.

wsanchez:
Install the IDNA tables file.

On trying to switch Klein to [use the Hypothesis strategies now provided by Hyperlink](https://github.com/twisted/klein/pull/393), I got [build failures](https://travis-ci.org/github/twisted/klein/jobs/722089054#L575) due to the file being missing.

I though that `tox -e packaging` would notice this was missing, but it doesn't catch it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     url=__url__,
     packages=find_packages(where="src"),
     package_dir={"": "src"},
-    package_data=dict(hyperlink=["py.typed"]),
+    package_data=dict(hyperlink=["py.typed", "idna-tables-properties.csv.gz"]),
     zip_safe=False,
     license=__license__,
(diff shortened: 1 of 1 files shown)


mahmoud (approved):
Weird that packaging doesn't catch this! Given that it doesn't, should we have `include_package_data` as a catch-all or do we want to keep with the remembering-to-explicitly-add?

wsanchez:
> Weird that packaging doesn't catch this!

Yeah I had the same thought…

> Given that it doesn't, should we have `include_package_data` as a catch-all or do we want to keep with the remembering-to-explicitly-add?

I'm not sure. It's rare, but now we need a release, which is annoying.  That said, I'd hate to package a thing we didn't mean to.  Dunno.

PR #135, opened 2020-08-28
Cap coverage version for requires.io

wsanchez:
Try moving extra comment out of `rq.filter` line to make `requires.io` happy


wsanchez:
Since this is trivial, doesn't change code, and has to merge to see if it works, I'm just gonna push it along…

PR #142, opened 2020-12-22
GitHub Actions CI

twm:
Travis is queuing builds for ages. 

I've also added Python 3.9 to CI and the package metadata. I notice that 2.6 isn't in CI. Should it still be listed as supported?

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -208,7 +208,7 @@ jobs:
           else  # CPython
               py_test="py${py/./}";  # Add "py" prefix, remove "."
           fi;
-          env_test="test-${py_test},codecov";
+          env_test="test-${py_test}";
           echo "Test environment: ${env_test}";
           ln -s ".tox/${env_test}" testenv;  # Fixed location for upload step below
           tox -e "${env_test}";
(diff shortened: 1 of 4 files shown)


wsanchez:
Hrm, I think we should add 2.6 to the matrix. That does seem like an oversight…
We bend over a bit to support 2.6 syntax, so it should be tested.

wsanchez:
I'm not seeing the workflows running… is that because this is an external PR and not a branch?

twm:
I honestly don't know. GHA is not great with the feedback... if you typo something in the YAML it doesn't tell you so it could be anything. I'll try creating a separate PR that's more obviously correct.

twm:
Nothing seems to be happening on #143, so I guess I can't configure GHA from my fork.

wsanchez:
@twm I'm getting ready for some travel this week but when I'm back in California, I can try making a branch with this and see if that works.

wsanchez:
@twm I added you as an admin to the repo, so you should be able to make a branch with this diff if you can get to it before me.

twm:
Thanks @wsanchez. I've recreated this PR over at #149. Closing this one in favor of that one, where GHA seems to be working.

PR #145, opened 2020-12-28
make hyperlink handle + like an HTML form post by default

### LINKED ISSUE #129: Hyperlink should treat + in query as %20
Via https://github.com/twisted/klein/issues/339, Hyperlink seems to be percent-encoding the `+` when parsing the URL:

```py
>>> from hyperlink import parse
>>> u = parse(b"https://localhost/getme?name=hello,%20big+world&value=4321".decode('ascii'))
>>> u
DecodedURL(url=URL.from_text('https://localhost/getme?name=hello,%20big%2Bworld&value=4321'))
>>> u.to_uri().to_text().encode('ascii')
b'https://localhost/getme?name=hello,%20big%2Bworld&value=4321'
```

This seems like a bug to me. `+` in the query part should be treated like `%20`. For example, this is what the [web API](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) does:

```js
u = new URL("https://localhost/getme?name=hello,%20big+world&value=4321");
console.log(u.searchParams.get("name"));
// outputs: hello, big world
```

glyph:
Per the *interminable* discussion already present there, fixes #129.

### DIFF
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -467,9 +467,13 @@ NO_NETLOC_SCHEMES = set(
 )
 # As of Mar 11, 2017, there were 44 netloc schemes, and 13 non-netloc
 
+NO_QUERY_PLUS_SCHEMES = set()
 
-def register_scheme(text, uses_netloc=True, default_port=None):
-    # type: (Text, bool, Optional[int]) -> None
+
+def register_scheme(
+    text, uses_netloc=True, default_port=None, query_plus_is_space=True
+):
+    # type: (Text, bool, Optional[int], bool) -> None
     """Registers new scheme information, resulting in correct port and
     slash behavior from the URL object. There are dozens of standard
     schemes preregistered, so this function is mostly meant for
@@ -485,6 +489,8 @@ def register_scheme(text, uses_netloc=True, default_port=None):
             not. Defaults to True.
         default_port: The default port, if any, for
             netloc-using schemes.
+        query_plus_is_space: If true, a "+" in the query string should be
+            decoded as a space by DecodedURL.
 
     .. _file an issue: https://github.com/mahmoud/hyperlink/issues
     """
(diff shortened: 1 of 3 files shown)


wsanchez (approved):
Looks good.

One question: do we really want to default to this for all schemes? I'm worried in particular about `urn:` schemes, though even in that case, perhaps the right answer is just "don't use `+` in URLs".

twm:
Unfortunately it's not just `DecodedURL` &mdash; `URL` also corrupts `+`. This PR demonstrates the issue: https://github.com/python-hyper/hyperlink/pull/146

twm:
:+1: Thank you for this! I'll update and release treq to depend on new Hyperlink version as soon as this has been released. That'll unblock Klein.

glyph:
> Looks good.
>
> One question: do we really want to default to this for all schemes? I'm worried in particular about `urn:` schemes, though even in that case, perhaps the right answer is just "don't use `+` in URLs".

The relevant specification for URNs seems to be https://tools.ietf.org/html/rfc8141#section-2.3.2 - previously, query parameters basically weren't used in URNs; https://tools.ietf.org/html/rfc2141 specifically punts on the issue. And the bananas new syntax in 8141 is going to require *extensive* changes in hyperlink to accommodate- for example, the URN equivalent of query components are now introduced by "?=" or "?+", not just "?".

In any event, URNs are not URLs so are in some sense out of scope for this library ;-).

I think if we want to handle these properly, we may want a new "URN" wrapper class which inherits this behavior and introduces the appropriate constraints (no netloc, for example) for URNs.

Given all this knock-on complexity, I'm going to go ahead and land it. I'm by no means opposed to supporting URNs, "q-component" and all, but we don't before this change and adding it won't change anything. (Arguably it will make it easier to build proper support on top of URL, although it will admittedly make DecodedURL a little more of a pain.)

PR #146, opened 2020-12-28
(Encoded)URL round-tripping of + in query

twm:
This demonstrates the failure afflicting https://github.com/twisted/treq/pull/304


glyph (approved):
Thanks for addressing the other aspect of this.

PR #149, opened 2021-01-05
GitHub Actions CI

twm:
Travis is queuing builds for ages.

I've also added Python 3.9 to CI and the package metadata. I notice that 2.6 isn't in CI. Should it still be listed as supported?

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -212,9 +212,7 @@ jobs:
           echo "Test environment: ${env_test}";
           tox -e "${env_test}";
           tar cvzf pytest-logs.tgz ".tox/${env_test}/log";
-          tox -e "coverage_combine-${env_test}";
-        env:
-          COVERAGE_FILE: .tox/coverage
+          tox -e "coverage_xml-${py_test}";
 
       - name: Upload pytest log artifact
         if: failure()
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -208,7 +208,7 @@ jobs:
           else  # CPython
               py_test="py${py/./}";  # Add "py" prefix, remove "."
           fi;
-          env_test="test-${py_test},codecov";
+          env_test="test-${py_test}";
           echo "Test environment: ${env_test}";
           ln -s ".tox/${env_test}" testenv;  # Fixed location for upload step below
           tox -e "${env_test}";
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -210,6 +210,7 @@ jobs:
           fi;
           env_test="test-${py_test}";
           echo "Test environment: ${env_test}";
+          tox -e "${env_test}";
           tar cvzf pytest-logs.tgz ".tox/${env_test}/log";
 
       - name: Upload pytest log artifact
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -1,0 +1,240 @@
+# Docs:
+# https://help.github.com/en/actions/automating-your-workflow-with-github-actions
+
+
+
+name: CI/CD
+
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
+
+
+jobs:
+
+  info:
+
+    name: Workflow information
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+
+    steps:
+
+      - name: Print GitHub Context
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "${GITHUB_CONTEXT}";
+
+      - name: Print Job Context
+        env:
+          JOB_CONTEXT: ${{ toJson(job) }}
+        run: echo "${JOB_CONTEXT}";
+
+      - name: Print Steps Context
+        env:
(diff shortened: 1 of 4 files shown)


twm:
All right! These are the diagnostically tractable failures I want in my life!

wsanchez:
Excellent

mahmoud:
Nice! 3.9 is of course fine, but it looks like most of the GHA jobs are still looking for 3.8. Do we want to change that, too? And I assume we're ok to remove traffic once GHA is working?

wsanchez:
> Nice! 3.9 is of course fine, but it looks like most of the GHA jobs are still looking for 3.8. Do we want to change that, too? And I assume we're ok to remove traffic once GHA is working?

Yeah I would run non-unit-test jobs to 3.9 and also toss Travis when GHA is working.

wsanchez:
GitHub CI doesn't include Python 3.4 or Python 2.6 (note that we don't have 2.6 running in Travis either).

Do we need to figure out how to run on 3.4, or are we OK with dropping it?  We could leave that on Travis, though it's a bit unfortunate to keep Travis in the loop just for that.

Similarly, if we aren't testing 2.6 in CI, should it remain on the supported list in `setup.py`?

wsanchez:
OMG ok, I think this is ready

wsanchez:
@mahmoud @glyph Remaining question is whether we need to continue testing 3.4 and whether the claim to support 2.6 is valid.  Both are going to increasingly create headaches for CI; we are already bending over considerably in order to make them work in `tox.ini`, but I've stopped installing either on any systems, so I, for one, am not testing them locally.

wsanchez:
Meanwhile, GHA has been done a while, but Travis hasn't even started building yet. (neither has Appveyor…)

wsanchez:
[GHA supports Windows (and macOS) also](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners), so we can probably nix Appveyor, but let's do that in another PR.

Kriechi:
I'd love to see the `cicd.yml` be a bit slimmer 😄 

With most of the other @python-hyper projects we outsource most of the heavy lifting straight to tox:
https://github.com/python-hyper/wsproto/blob/master/.github/workflows/ci.yml
https://github.com/python-hyper/hyper-h2/blob/master/.github/workflows/ci.yml
https://github.com/python-hyper/hyperframe/blob/master/.github/workflows/ci.yml
https://github.com/python-hyper/hpack/blob/master/.github/workflows/ci.yml

twm:
@Kriechi Is there some tooling for keeping that in sync, or otherwise an advantage to aligning with the rest of the org? I was trying to imitate Klein's config (read: I copied it) for a few reasons:

* I knew @wsanchez would be involved in review so, so copying a config he is familiar with it would ease that process
* The Klein config is actively maintained, so improvements can easily flow from it to a derived config, even if that process is manual

Kriechi:
@twm no automation or tooling - just personal preference to keep it in sync across the org. It would allow other maintainers within the org to maintain this project as well.

No strong reasons as you can see - just a nice-to-have IMO.

wsanchez:
@Kriechi I appreciate the desire to make things consistent, but the developers are not consistent.

Having separate jobs in the CI config allows GHA to display more granular information about the builds, and hopefully someday, to restart specific jobs within a run, so the strategy of running all tox environments in one job seems pretty flawed to me.

wsanchez:
@twm can you give the changes I made a once-over?  I feel weird marking this as reviewed since I also made changes, so would like a double-check.

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,238 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install Tox
+        run: pip install tox;
+
+      - name: Run unit tests
+        shell: bash
+        # This hairy shell code is here to map the Python versions
```
twm, line 201: You may appreciate how I did this in treq:

https://github.com/twisted/treq/blob/349c44e3ef5f52e635e56bb54727af63efdbeafc/.github/workflows/ci.yaml#L95-L110

twm (commented):
@wsanchez This looks good to me!

glyph (approved):
1. I endorse @twm's review. He is good at reviewing and good at CI.
2. Gosh it is [so depressing that every project needs to write this enormous YAML file by hand](https://glyph.twistedmatrix.com/2020/02/modules-for-maintenance.html)
3. Travis is awful so I'm just gonna merge this now. If there are issues we can always back it out or update it; @wsanchez I'll let you integrate @twm's suggestion at your leisure.

.github/workflows/cicd.yml
wsanchez: OMG yeah I like that better. Will revisit when I get some more cycles for that.

wsanchez:
> * Gosh it is [so depressing that every project needs to write this enormous YAML file by hand](https://glyph.twistedmatrix.com/2020/02/modules-for-maintenance.html)

I started thinking about a project in which you would have Yet Another Config File that described at least the things that are different between the projects I have my fingers in for now, and it will generate a (probably larger than optional but deterministic) Tox file and GHA CI file and maybe other such things that all play nice together, because I know try to keep about a dozen of these things in sync manually.  But it's just an idea and no code yet.

PR #157, opened 2021-03-16
Fix some small type issues.

### LINKED ISSUE #156: Small type issues
Hi.  Thanks for adding types to hyperlink!  I noticed 2 smallish improvements that could be made.

1. It would be nice if `parse_url` used `typing.overload` to determine which type of URL it returns (i.e. DecodedURL vs URL).
2. I believe QueryParameters could just be typed as:
```
QueryParameters = Iterable[Tuple[Text, Optional[Text]]
```
(This way it can also accepts generators and comprehensions)

Let me know if you'd like me to make a PR with these changes.

euresti:
Use @overload to have a better return type for parse.
Use Iterable for QueryParameters to support generators and views.

Fixes #156


euresti:
I think the appveyor errors are caused by pip no longer supporting 2.7 nor 3.5.

src/hyperlink/_url.py
```diff
@@ -2415,6 +2420,25 @@ def __dir__(self):
     # # End Twisted Compat Code
 
 
+# Add some overloads so that parse gives a better return value.
+@overload
+def parse(url, decoded, lazy=False):
+    # type: (Text, Literal[False], bool) -> URL
```
wsanchez, line 2426: neat-o

wsanchez (approved):
Awesome, thanks

wsanchez:
@mahmoud @glyph Any objection to dropping 2.7 and 3.5 on Windows from the build matrix?

glyph:
None from me! Into the memory hole.

PR #179, opened 2023-01-31
Fix non-numeric port number bug (addresses #180)

kenballus:
This isn't supposed to work:
```python3
import hyperlink
hyperlink.URL.from_text("http://example.com:-80")
```
This patch ensures that only valid port strings are accepted.

(The root cause of this is that port numbers are parsed using `int`. Both cpython and rfc3986 also had this bug.)

### DIFF
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -478,7 +478,7 @@ NO_QUERY_PLUS_SCHEMES = set()
 def register_scheme(
     text, uses_netloc=True, default_port=None, query_plus_is_space=True
 ):
-    # type: (Text, bool, Optional[int], bool) -> None
+    # type: (Text, bool, Optional[Union[int, str]], bool) -> None
     """Registers new scheme information, resulting in correct port and
     slash behavior from the URL object. There are dozens of standard
     schemes preregistered, so this function is mostly meant for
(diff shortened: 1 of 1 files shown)


kenballus:
Addresses #180

src/hyperlink/_url.py
```diff
@@ -501,9 +501,9 @@ def register_scheme(
     """
     text = text.lower()
     if default_port is not None:
-        try:
+        if isinstance(default_port, int) or (isinstance(default_port, str) and default_port.isdigit() and default_port.isascii()):
```
mahmoud: Seems like someone could register a negative default port here? Since registering is not performance sensitive in the slightest, we could wrap it in a `str()` and run the same check for consistency?

mahmoud (commented):
Nice work! Just one question/thought for now. Thanks for this!

src/hyperlink/_url.py
kenballus: Good catch!
I just pushed up a commit that does exactly what you suggest.
(It also removes some calls to functions not present in python 2, and updates comments/errors)

mahmoud (approved):
LGTM, thanks for addressing the feedback so quickly!

kenballus:
Any update on this? Thanks!

Issue #188, opened 2026-01-30
URL.to_uri() applies IDNA encoding to IPv6 literal hosts 


mvaught02:
### Description

When calling `URL.to_uri()` on a URL that uses a **plain IPv6 literal host** (e.g. `http://[::1]:PORT/...`), `hyperlink` attempts to IDNA-encode the host. This is incorrect and results in an exception from `idna`, since IPv6 literals are **not DNS labels** and must not undergo IDNA processing.

This is the same root cause as #68, which appears to have gone stale.

---

### Reproduction

```python
from hyperlink import EncodedURL

url = EncodedURL.from_text("http://[::1]:60993/api/v1/token")
url.to_uri().to_text().encode("ascii")
```

### Actual result

```
idna.core.InvalidCodepoint:
Codepoint U+003A ':' not allowed
```

This occurs because `idna.encode("::1")` is invoked.

---

### Expected result

The URL should round-trip successfully:

```python
"http://[::1]:60993/api/v1/token"
```

---

### Why this is a bug

- IDNA applies **only to DNS labels**, not IP literals
- RFC 3986 explicitly allows IPv6 literal hosts
- `hyperlink` already has `parse_host()` which distinguishes:
  - DNS names
  - IPv4 literals
  - IPv6 literals
- Applying IDNA to IP literals is both unnecessary and invalid

---

### Suggested fix

In `URL.to_uri()`, detect IPv4/IPv6 literal hosts using `parse_host()` and **skip IDNA encoding** for those cases. IDNA should only be applied to non-IP (DNS) hosts.

Pseudo-logic:

```python
family, host_text = parse_host(self.host)
if family is not None:
    # IPv4 / IPv6 literal
    host = host_text
else:
    host = idna_encode(host_text, uts46=True).decode("ascii")
```

---

### Regression test

```python
def test_ipv6_literal_not_idna_encoded():
    from hyperlink import EncodedURL

    url = EncodedURL.from_text("http://[::1]:8080/path")
    assert url.to_uri().to_text() == "http://[::1]:8080/path"
```

---

### Environment

- hyperlink: current release / main
- Python: 3.10+
- idna: current

Issue #68, opened 2018-12-05
URLs don't support fromText -> toURI with URLs containing IPv6 literals


hawkowl:
```
>>> URL.fromText(u"http://[3fff::1]/foo").asURI().asText()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/hawkowl/venvs/commands/lib/python2.7/site-packages/hyperlink/_url.py", line 1338, in to_uri
    new_host = self.host if not self.host else idna_encode(self.host, uts46=True).decode("ascii")
  File "/home/hawkowl/venvs/commands/lib/python2.7/site-packages/idna/core.py", line 340, in encode
    s = uts46_remap(s, std3_rules, transitional)
  File "/home/hawkowl/venvs/commands/lib/python2.7/site-packages/idna/core.py", line 332, in uts46_remap
    _unot(code_point), pos + 1, repr(domain)))
idna.core.InvalidCodepoint: Codepoint U+003A not allowed at position 5 in u'3fff::1'
```

mahmoud:
Hey Hawkie! This was pretty concerning at first, since I thought we had a bunch of ipv6 coverage, but now I see, so the problem is actually the `to_uri()` part and the newly-integrated idna stuff:

```
>>> url = URL.from_text(u'https://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:80/')
>>> url.to_uri()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mahmoud/virtualenvs/tmp-d364b3d6b21cd4e4/local/lib/python2.7/site-packages/hyperlink/_url.py", line 1338, in to_uri
    new_host = self.host if not self.host else idna_encode(self.host, uts46=True).decode("ascii")
  File "/home/mahmoud/virtualenvs/tmp-d364b3d6b21cd4e4/local/lib/python2.7/site-packages/idna/core.py", line 358, in encode
    s = alabel(label)
  File "/home/mahmoud/virtualenvs/tmp-d364b3d6b21cd4e4/local/lib/python2.7/site-packages/idna/core.py", line 270, in alabel
    ulabel(label)
  File "/home/mahmoud/virtualenvs/tmp-d364b3d6b21cd4e4/local/lib/python2.7/site-packages/idna/core.py", line 304, in ulabel
    check_label(label)
[...]

PR #189, opened 2026-01-30
Do not IDNA-encode IPv6 literal hosts in to_uri() (#188)

mvaught02:
### Summary

This PR fixes a long-standing bug in `hyperlink.URL.to_uri()` where plain IPv6 literal hosts (e.g., [::1]) are incorrectly passed to `idna.encode()`.

This bug was originally reported in #68 and #188 and affects downstream users (e.g., treq) even on current releases.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -94,6 +94,7 @@ skip_install = True
 
 deps =
     black==21.7b0
+    click<8.0
 
 setenv =
     BLACK_LINT_ARGS=--check
--- a/src/hyperlink/_url.py
+++ b/src/hyperlink/_url.py
@@ -1667,11 +1667,16 @@ class URL(object):
         new_path = _encode_path_parts(
             self.path, has_scheme=bool(self.scheme), rooted=False, maximal=True
         )
-        new_host = (
-            self.host
-            if not self.host
-            else idna_encode(self.host, uts46=True).decode("ascii")
-        )
+        family, host_text = parse_host(self.host)
+        if family is not None:
+            # IPv4 / IPv6 literal
+            new_host = host_text
+        else:
+            new_host = (
+                self.host
+                if not self.host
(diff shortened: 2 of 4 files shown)


mvaught02:
@wsanchez / @glyph can you give this one a look?

glyph:
ugh, looks like our CI needs some updates first, to catch up with more recent pythons and linuxes

mvaught02:
Hmm, I can probably tackle this a couple of ways.

When I opened this PR, I was trying to keep it low-touch in hopes we could get a quick release with the fix without dragging in a bunch of unrelated changes. I can stick with that theme and just “make it work” — update the Actions config to use a modern Python for the checks and switch the legacy test versions over to containers so they still run. I’m happy to wire that up in this PR.

Or, if it makes more sense, I can either expand this PR to uplift the project to currently supported Python versions, or open a separate PR to handle that cleanup first and then we can merge this one afterward.

I’m good either way — just let me know what you’d prefer and I can take care of it. At a glance this doesn’t seem like a major lift. Mostly just trying to get this fix in since it keeps coming back to annoy me 😅

glyph:
> Or, if it makes more sense, I can either expand this PR to uplift the project to currently supported Python versions, or open a separate PR to handle that cleanup first and then we can merge this one afterward.

A separate PR for CI cleanup would be ideal. But this change is relatively small and I *think* uncontroversial so if you preferred to clean it up here I wouldn't complain _too_ much.

tox.ini
```diff
@@ -97,6 +97,7 @@ skip_install = True
 
 deps =
     black==21.7b0
+    click<8.0
```
glyph, line 100: What's this bound about?

mvaught02: When I ran this locally for tests I had to pin this back to get tests to work with older versions. This goes away when that other PR is merged.

Issue #185, opened 2025-08-18
Dropping Python 2


mahmoud:
Hey folks! Was looking at using hyperlink today and I saw we're a bit behind on supporting the newer Python versions. I started a branch, but ran into trouble getting everything working with Python 2. (The tox even has py26.) The tox.ini is a thing of beauty, one of the most organized I've seen. But how do we feel about dropping Python 2 and getting some new versions in there, @wsanchez / @glyph ?

glyph:
Let's drop it. I realize we want to be a *bit* more lenient given that this is a very core-infrastructure sort of thing, but I think 3.6 is the most generous floor by now. 2.x is just silly at this point.

PR #190, opened 2026-02-26
Update to latest supported python versions

mvaught02:
Addresses #185 and uplift discussion from #189 

Summary of Changes:

   - Drop Python 2 Support: Removed all Python 2 compatibility code (six, __future__ imports related to py2, PY2 conditionals, unicode vs str logic).
   - Modernize Type Hints: Converted comment-based type hints to inline Python 3 type hints across the codebase.
   - Updated Imports: Moved typing-only imports to if TYPE_CHECKING: blocks and used from __future__ import annotations to avoid runtime overhead and circular imports as well as appease the latest lint rules.
   - Update Supported Versions: Updated setup.py, tox.ini, and docs/index.rst to support Python 3.10–3.14. Removed pypy (python2) from default test envs.
   - Remove Legacy Code: Deleted src/hyperlink/_socket.py (ctypes fallback for inet_pton) and src/hyperlink/test/test_socket.py.
   - Fix Hypothesis Strategy: Fixed a bug in hostname_labels strategy where generated IDNs could exceed the maximum allowed length in Punycode, causing test failures.

Open questions...

do we want to go further back to older python versions? doing so will require using containers to get old python versions (github actions only supports 3.10-3.14) but if we want to take this further back from a test perspective we could if thats critical from a code perspective i think this would work all the way back to 3.7+


mvaught02:
@glyph sorry it took me a little bit to get back to this but went ahead and did a separate PR and just did the uplift and the py2 stuff here

glyph (approved):
This all looks good to me. In the future, though, please submit smaller PRs, so there's less code to look at in one big pile. Luckily most of these changes are mechanical so it wasn't *that* bad though — hopefully I didn't miss anything sneakily hiding in between the obvious `Text` → `str` replacements 😅

glyph:
sigh, looks like we need to upgrade `upload-artifact` for this to work

mvaught02:
> This all looks good to me. In the future, though, please submit smaller PRs, so there's less code to look at in one big pile. Luckily most of these changes are mechanical so it wasn't *that* bad though — hopefully I didn't miss anything sneakily hiding in between the obvious `Text` → `str` replacements 😅 

Yes I'm sorry it was larger than I had wanted there was not a good way to break this up that I had seen originally. 😅

mvaught02:
> sigh, looks like we need to upgrade `upload-artifact` for this to work

I'll check this here shortly and get it updated.

